PR #20, opened 2016-04-02
remove unused t from tele_var_t and tele_array_t

samdoshi:
I managed to get [flycheck][] working with properly with emacs and avr32, anyway it's flagged a few things... in particular a bunch of uninitialized fields.

So, as far as I can tell the field `t` is not used in `tele_var_t` and `tele_array_t`, I'm guessing it's left over from some previous use?

I haven't tested this change on hardware...

[flycheck]: http://www.flycheck.org/


tehn:
thanks!

PR #24, opened 2016-04-05
.editorconfig & .clang-format

samdoshi:
I've added a `.editorconfig`([website](http://editorconfig.org/)) and a `.clang-format` ([website](http://clang.llvm.org/docs/ClangFormat.html)) to the `src` directory.

I've re-formatted `config.mk` by hand.

I've added a new `make` target `format` (it's not the best bit of Makefile magic you'll ever see...), and run it to reformat the teletype source code.

If you don't particularly like any of the choices I've made it `.editorconfig`  or `.clang-format`, let me know and I can rerun.


src/table.c
```diff
@@ -4,60 +4,90 @@
...
+        21299, 21572, 21845, 22118, 22391, 22665, 22938, 23211, 23484, 23757,
+24030, 24303,
+        24576, 24849, 25122, 25395, 25668, 25941, 26214, 26487, 26761, 27034,
+27307, 27580,
+        27853, 28126, 28399, 28672, 28945, 29218, 29491, 29764, 30037, 30310,
+30583, 30857,
+        31130, 31403, 31676, 31949, 32222, 32495, 32768, 33041, 33314, 33587,
+33860, 34133,
+        34406, 34679
 };
```
pq: This looks odd to me. Maybe a matter of setting `ReflowComments` to false?

src/table.c
```diff
@@ -4,60 +4,90 @@
...
+    2827,  2897,  2967,  3039,  3112,  3186,  3262,  3339,  3417,  3497,  3578,
+    3660,  3744,  3829,  3916,  4005,  4094,  4185,  4278,  4373,  4468,  4566,
+    4665,  4765,  4868,  4971,  5077,  5184,  5293,  5403,  5516,  5630,  5745,
+    5863,  5982,  6104,  6227,  6351,  6478,  6607,  6737,  6870,  7004,  7140,
+    7279,  7419,  7561,  7706,  7852,  8000,  8151,  8304,  8459,  8616,  8775,
+    8936,  9100,  9266,  9434,  9604,  9777,  9952,  10129, 10309, 10491, 10675,
+    10862, 11051, 11243, 11437, 11634, 11833, 12035, 12240, 12447, 12656, 12869,
+    13083, 13301, 13521, 13744, 13970, 14199, 14430, 14664, 14901, 15141, 15384,
+    15629, 15878, 16129
 };
```
pq: Can clang insert a newline at EOF or is that an editor config thing?

pq: Ah, wait, I see it in `.editorconfig`.  Maybe this file just needs to be reprocessed.

src/table.c
samdoshi: It's possible that this just comment block just needs deleting. @tehn?

src/table.c
samdoshi: .editorconfig only applies on editor save. Not when clang-format runs. I'll edit it and force push.

src/Makefile
```diff
@@ -64,3 +64,13 @@
...
+
+.PHONY: format
+format:
+	clang-format -style=file -i main.c
+	clang-format -style=file -i table.h
+	clang-format -style=file -i table.c
+	clang-format -style=file -i teletype.h
+	clang-format -style=file -i teletype.c
+	clang-format -style=file -i euclidean/euclidean.h
+	clang-format -style=file -i euclidean/euclidean.c
```
pq: What about something like 
```
clang-format -style=file -i *.c *.h
clang-format -style=file -i euclidean/*.c euclidean/*.h
```

Annoying that `**/*.c` doesn't work. 

**EDIT:** better still?:

```
find . -name '*.c' -o -name '*.h' | xargs clang-format -style=file -i
```

samdoshi: Problem is we don't necessarily want to format all the files. e.g. `euclidean/data.*` is auto generated by `euclidean.hs`. Not that it really matters that much...

pq: Aha!

Well, you're right that it probably doesn't matter...  But if it does maybe add a ``no format`` comment to the header of the file?

samdoshi: It's a good idea. At some point (hopefully in the not too distant future) I'd like to split the code up into smaller files, so some sort of globing will be necessary. I'll probably modify the Haskell code to generate the files as you suggest.

pq: Hadn't noticed that you're generating from Haskell. Neat!

Another consideration would be to add codegen to the build process and then not commit the generated C sources.  Of course, that requires all devs have haskell installed...  (Maybe that's a show-stopper.)

But anyway, cool!

src/table.c
tehn: yes let's throw it away

Issue #28, opened 2016-04-06
Setup travis build.


pq:
A few bits:

- [x] setup `.travis.ym;` appropriately --- @samdoshi has proved this out (https://github.com/samdoshi/teletype/commit/0a51dd8c1efd6fc3865cc624b80d52108b2ec5af), so this shouldn't be too hard
- [x] add a build badge to the `README`

pq:
@tehn : looks like an admin needs to flip the bit on travis for teletype: 

![screen shot 2016-04-06 at 2 33 38 pm](https://cloud.githubusercontent.com/assets/67586/14333624/9900ad62-fc04-11e5-8d42-d97511c47598.png)

https://travis-ci.org/profile/monome

Could you?

tehn:
flipped

tehn:
i must say, i am completely baffled by the contents of .travis.yml

pq:
Ha!  I blame @samdoshi.

tehn:
ah, it makes more sense now that i actually read about travis.

so right now we're just doing a compilation check, correct?

pq:
> so right now we're just doing a compilation check, correct?

Right. Once we have tests we should add them.  Down the road we could add linting or some other kind of presubmit checks.  Note that pull requests trigger a build which means you can see if a PR will breaj the build before accepting it.

For example:

![screen shot 2016-04-06 at 4 02 27 pm](https://cloud.githubusercontent.com/assets/67586/14335552/fbc1808c-fc10-11e5-8935-6bdeddff313a.png)

pq:
Added badge with https://github.com/monome/teletype/commit/cbb9fe8eb7613dfa91ccdc9d16a9953af4165b02 which means this first bit is done.

@samdoshi or @tehn : feel free to re-open or create another issue if you want to see refinements to the travis build.

tehn:
looks good, thanks!

samdoshi:
> i must say, i am completely baffled by the contents of .travis.yml

You and me both.

Should we get https://github.com/samdoshi/avr32-toolchain-linux move over too?

pq:
> Should we get https://github.com/samdoshi/avr32-toolchain-linux moved over to the 'monome' org too?

:+1: 

Make sense to me!

PR #29, opened 2016-04-06
Travis config first cut.

pq:
Cribbed from https://github.com/samdoshi/teletype/commit/0a51dd8c1efd6fc3865cc624b80d52108b2ec5af.

Feels dodgy to use Sam's avr32-gnu-toolchain zip but maybe sufficient for a first cut?

I also think we should update the `language` to C (and if necessary the build bits) in another pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+language: generic
+
+install:
+  - pushd $HOME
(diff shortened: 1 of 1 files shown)


pq:
cc @tehn @samdoshi

pq:
First part of https://github.com/monome/teletype/issues/28.

PR #32, opened 2016-04-07
Unit tests

samdoshi:
Adds unit tests using the [greatest](https://github.com/silentbicycle/greatest) library.

Moves the simulator up a directory

Also commit 7c66926 changes the prototype for `parse` so as to remove the global `temp` variable.


pq:
Awesome.

So nice too that you added a `make test` target.  Behold:

screen shot 2016-04-07 at 9 12 19 am

Moving forward, having automated tests to vet PRs should be a real boon.

Cheers!

.travis.yml
```diff
@@ -1,4 +1,4 @@
-language: generic
+language: c
```
pq: Yay!

Do we want to specify compiler?  (Or have a matrix?)

For example:

```
compiler:
  - clang
  - gcc
```
(From: https://docs.travis-ci.com/user/languages/c)

samdoshi: I made it `generic` initially as we bring our own toolchain!

Multiple compilers seems a bit overkill to me, for now the only target that matters is the hardware (really we should install gcc-4.4, but baby steps...).

pq: Sounds good.  If gcc is the expectation may not hurt to specify it but no biggie either way!

simulator/.clang-format
```diff
@@ -0,0 +1,30 @@
...
+  AfterUnion:            false
+  BeforeCatch:           false
+  BeforeElse:            true
+  IndentBraces:          false
+
+Cpp11BracedListStyle: false
+
+MaxEmptyLinesToKeep: 2
+---
+
```
pq: Is there a reason not to put these at the root dir?  IOW do we not want to format the other bits too?

simulator/.editorconfig
```diff
@@ -0,0 +1,22 @@
...
+[{Makefile,config.mk}]
+indent_style = tab
+indent_size = 8
+trim_trailing_whitespace = true
+
+# C & Cpp
+[*.{h,hpp,c,cpp}]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
```
pq: Same question here.

simulator/.clang-format
samdoshi: Basically, yes, there is the libavr32 submodule which is used by other modules. 

I'm just being cautious, if everyone gets on with using an editorconfig and clang-format then we'll do libavr32 as well, and then move these files to the root.

pq: Oh, right.  Cool.  Thanks for the clarification!

PR #40, opened 2016-04-23
Cleanup teletype dep on avr.

pq:
Can we do without this include?

`print_funcs.h` pulls in a bunch of avr bits that are not needed/wanted when running teletype off the device.

cc  @samdoshi @tehn

### DIFF
--- a/src/teletype.c
+++ b/src/teletype.c
@@ -14,7 +14,6 @@
 #ifdef SIM
 #define DBG printf("%s", dbg);
 #else
-#include "print_funcs.h"
 #define DBG print_dbg(dbg);
 #endif
 


samdoshi:
Why not just pass `-DSIM` to your compiler? That will remove the include.

pq:
> Why not just pass -DSIM to your compiler? That will remove the include.

Because I'm an XCode newbie and I'm just trying to get it to stop yelling at me? ;)

I'm guessing the non sim case breaks otherwise?  If that's the case, I'd expect that include to happen in `main.c`?

pq:
Popping up, I'm curious why the include is there but can certainly work around it with `-DSIM`.  Your call!

samdoshi:
As far as I know the include is used to making debugging over the serial port easier. It won't be broken as normally there aren't any debug statements uncommented. (And I deleted a lot when I went through the code.)

It's not a particularly clean solution, but I think if you want to remove that include, you'll need to make the rest of it consistent (i.e. deal with `DBG`), or better yet come up with an entirely different solution to conditional debugging.

Just incase you haven't got it working yet: http://stackoverflow.com/questions/26928622/add-preprocessor-macro-to-a-target-in-xcode-6

tehn:
yeah, we need that in there. a SIM define should solve your issues.

pq:
Thanks guys.  @tehn: is this just as Sam suggests in order to make debugging on the device work if you uncomment debug statements?  Or are other bits flipped/defined by the include too?  Just trying to understand how it all hangs together!

tehn:
the include simply redefines the print_dbg() functions so they get directed
at the avr UART rather than to stdio (screen)

so the idea is you can use the same debug prints for either the SIM or the
AVR.

theoretically if you were making a max external you could do a define so
that print_dbg redirects to printing at the max window, however you do that

pq:
Thanks @tehn.  I'll stick to the `SIM` define for now.  Cheers!

PR #50, opened 2016-10-11
Orca + JF tweaks

samdoshi:
I've renamed the Orca functions to match the name of the ops.

Also added the `NOTUSED` macro to the JF functions.


scanner-darkly:
:+1:

PR #60, opened 2017-01-03
protect tele_ii_tx from interrupts

ngwese:
**_note: this change is untested and intended primarily for review._**

no idea (yet) if this helps with stability/lockups but while working through the ii code to understand how it all works i noticed that the queue manipulation in ```tele_ii_tx()``` isn't protected from changes made to the queue (at interrupt level) in ```handler_IItx()```.

this change:
- disables interrupts while filling a queue entry in order to prevent a partially filled queue entry from being transmitted
- mild reordering of ```waiting``` flag handling to reduce potential for partially filled ii queue entries being marked as sendable or marked as available before being sent.
- guards against a potentially infinite loop/memory corruption case when searching for an available ii queue entry to pack (in ```tele_ii_tx()```).

### DIFF
--- a/module/main.c
+++ b/module/main.c
@@ -51,6 +51,7 @@ http://msgpack.org
 #include "timers.h"
 #include "util.h"
 // #include "ftdi.h"
+#include "conf_tc_irq.h"
 #include "font.h"
 #include "hid.h"
 #include "region.h"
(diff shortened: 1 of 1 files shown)


ngwese:
....after examining the code more i don't think some of the above changes have any real effect so beyond the protection against an infinite loop and the volatile change the reset is not needed (since the normal event queue provides most of the needed safety).

i'm going to restructure this PR.

PR #70, opened 2017-02-24
LV.CV | CY.CV implementation

scanner-darkly:
implementation for `LV.CV` and `CY.CV` remote commands
- these will return value at the specified CV output for Levels and Cycles respectively
- as discussed these are only readable for now
- ansible will return 0 for CV number outside of 1..4 range


scanner-darkly:
@tehn i left the version number the same as for telex, figured the changes could be packed into the same release - let me know if i should change it (might be a good idea to make version numbers more granular?)

PR #90, opened 2017-09-04
Fixes #38 by making AVG and Q.AVG round UP

### LINKED ISSUE #38: Behaviour of AVG and Q.AVG?
At the moment `AVG` and `Q.AVG` perform integer division when calculating the average.

e.g.

```
AVG 2 3
>>> 2
```

Should they round instead?

burnsauce:
AVG -32767 -32768 == -32767


burnsauce:
Does anyone prefer "round away from zero" instead of "round up"?

PR #92, opened 2017-09-05
New Op: W (while loop)

burnsauce:
e.g.: W LT A 500: A * A B

Maximum depth set by WHILE_DEPTH (default 10000)


samdoshi:
This is going to need some serious thought as up till now the language has never allowed the re-execution of commands unless they are after `:`.

I need to make sure that there aren't future consequences of it.

Issue #94, opened 2017-09-05
Each execution state needs an I variable


burnsauce:
```
1:
L 1 4: SCRIPT 2; A I
2:
L 5 8: B I

A -> 8
B -> 8
```

I solved this problem with the while loop by tracking per execution level in es_state, which made sense but was not ideal.  I is in ss.variables. Does `ss.variables.i[EXEC_DEPTH]` offend anyone?  If not, I can implement a fix that way.  I'm amenable to some refactoring if someone sees an abstraction that works better.

samdoshi:
Counterpoint (and untested on hardware):

```
1:
L 1 4: SCRIPT 2; A I

2:
I 100

A -> 100
```

I would expect `A` to be `100` after executing script `1`. With the change you are suggesting it would be just `4` wouldn't it?

burnsauce:
The nature of a for loop suggests that I should contain the iterator of the
current loop.  As it stands, calling a script with a loop from a loop
doesn't work as expected.  In other words, nested loops need their own
iterators.

This is the corner case of L and SCRIPT.  W and SCRIPT absolutely required
per exec depth tracking.  I think it makes sense for L, too.

burnsauce:
Come to think of it, won't IF and SCRIPT present the same corner case?

samdoshi:
I still disagree with regards to `I`, plus it will be a breaking change.

Also, can you explain the issue with `IF` and `SCRIPT` for me.

burnsauce:
Not at the hardware, but I expect this to misbehave:

1:
A 0
IF EZ A: SCRIPT 2
ELSE: A 1

2:
IF NZ A: A 2

I think A will be 1.

samdoshi:
Gotcha.

I'm wondering if we should instead think about pushing and popping the execution state then? And track the execution depth via the height of the stack.

Or maybe keeping `exec_state_t` for the entire execution context, but having a member for things that need pushing and popping via `SCRIPT`:

e.g.

```c
typedef struct {
    bool if_else_condition;
} nested_exec_depth_t;  // naming things is hard.

typedef struct {
    nested_exec_depth_t nested[EXEC_DEPTH];
    uint8_t exec_depth;
} exec_state_t;
```

Plus high level functions in `state.c` to push, pop, and obtain the depth.

burnsauce:
I think a stack is appropriate.  I'll bang it out tomorrow afternoon.

samdoshi:
[...]

PR #96, opened 2017-09-07
Fixes Issue #94 (breaking existing IF and I behaviour)

burnsauce:


burnsauce:
FYI, this will break the Pull requests for the W and BREAK operators (PR91 and PR92), as they also require the use of per-execution-state variables.

If you merge this one before the other two, I'll patch them up.  (I'm always willing to do the work patching my PRs if they fail to merge cleanly with master.)

burnsauce:
I can update this patch to provide less-breaking behaviour by carrying the last-updated `I` back to ss.variables.i and modifying the `I` op to return that value whenever it's not being called from within a subcommand of mod `L`.

It will always be breaking in a patch that reads or writes I from within an L loop that calls SCRIPT.  I don't know that anyone does in a manner that relies on this broken corner case but I don't think the existing behaviour is sane or desirable.

PR #103, opened 2017-09-22
Add a PR template

cmcavoy:
Github provides the ability to load pull request templates when a new PR is submitted. The template lives in either the `.github` or `docs` directory. To keep `docs` clean, I created a `.github` directory and added the template there. The PR template aligns with the discussion organization effort on lines.

More information on Github support [here](https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/).

Here's this PR using the template,

#### What does this PR do?
Github provides the ability to load pull request templates when a new PR is submitted. The template lives in either the `.github` or `docs` directory. To keep `docs` clean, I created a `.github` directory and added the template there. 

More information on Github support [here](https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/).

#### Any background context you want to provide?
The PR template aligns with the discussion organization effort on lines.

PR #105, opened 2017-09-28
New Issue Template

cmcavoy:
#### What does this PR do?
Adds a template for new Github issues.

It was suggested in this [PR](https://github.com/monome/teletype/pull/103) by @samdoshi 

Only way to test it is to merge it. I've checked the markdown formatting in Atom Github-flavor markdown preview and it worked ok.

I'm concerned this is too heavy weight and might intimidate new issue writers. The PR is editable by maintainers, so I'd suggest making edits directly rather than waiting for me to edit it based on feedback. Think of this PR as an initial draft to get thoughts going.

#### I have,
* [ ] updated CHANGELOG.md

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,6 +1,6 @@
 #### Please describe the bug.
 
-#### What release of the firmware are you using? Please include the hash displayed at startup and the release of the software.
+#### Please include the hash displayed at startup and the release of the software (e.g. "Teletype v2.0.1 5f838c9")
 
 #### What steps are required to reproduce the bug?
 1. step 1
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,8 @@
+#### Please describe the bug.
+
+#### What version of the firmware are you using?
+
+#### What steps are required to reproduce the bug?
+1. step 1
+1. step 2
+1. step 3
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,6 +1,6 @@
 #### Please describe the bug.
 
-#### What version of the firmware are you using?
+#### What release of the firmware are you using? Please include the hash displayed at startup and the release of the software.
 
 #### What steps are required to reproduce the bug?
 1. step 1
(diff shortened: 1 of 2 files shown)


samdoshi:
The bug template should have a "version reported on" field, in particular, asking people to include the hexadecimal string at the end of the version number (it's the git hash). Probably best to include some text explaining that the version number is displayed at power on too.

I'm not 100% convinced by the need to have 2 types of templates, I think just the bug one would be best, and really just as a way to "nudge" people to include relevant info.

If we do keep the "feature template" part, can we tone down the corporate speak a bit, I think it sends the wrong kind of signal. I think it's just the "acceptance criteria" section that's guilty of this, it might be easiest to just get rid of it?

(Thank you for taking point on this!)

samdoshi:
Also, I hope you don't mind me mentioning it... (but as you're the one responsible for this kind of stuff now...)

The first line of a commit message shouldn't be too long as it messes with the output of `git log` amongst other things. Generally it's short one line summary, followed by a blank line, and then followed by a longer summary (if needed).

If you Google "how to write a commit message" there are a few articles out there, most of which specify 50 chars for the first line, which is probably a bit over the top for us.

cmcavoy:
> Also, I hope you don't mind me mentioning it

Nope, participating in the open means being open to feedback. All is well.

> 50 chars for the first line

Noted, will follow those guidelines on future commits. Thanks!

scanner-darkly:
+1 for version but should probably include a note saying for TT it's what's displayed at bootup, and for the rest it should be the release version (which necessitates us providing some way to identify betas as well - i've been including the commit # when posting beta versions for this purpose).

cmcavoy:
Updated to include field for version number / bootup hash.

samdoshi:
Is it wise to use the term 'hash', no not for that reason... I mean that some people may have no idea what it means. Not sure what the best alternative is. Maybe just give an example, e.g.

```
#### What release of the firmware are you using? 
Please include the hash displayed at startup and the release of 
the software (e.g. "Teletype v2.0.1 5f838c9")
```

cmcavoy:
@samdoshi I think your second line says it perfectly, "Please include the hash displayed at startup and the release of the software (e.g. "Teletype v2.0.1 5f838c9"). I bumped that up to the field, rather as a supplement / help text.

cmcavoy:
I forgot I've got the merge button now. If no one objects in the next 24 hours or so, I'll merge this in. /cc @tehn @samdoshi @scanner-darkly @burnsauce (let me know if I should be cc'ing anyone else here.)

samdoshi:
Sounds good to me. Go with the "squash and merge" option though.

PR #113, opened 2017-10-09
Fixes #108 ~ key docs bug

### LINKED ISSUE #108: Tilde key does not open TRACKER
The documentation states that the tilde (~) key will open tracker mode, but it doesn't.  Instead, it displays the £ character.

1. Press the tilde key

jlmitch5:
Straightforward docs change...relevant to >= 2.0.0 releases.

Figured it was easiest to just commit to master and get this into 2.1.0 rather than try to backport to 2.0.0/2.0.1.


samdoshi:
Is it worth ditching the name TRACKER too? I'm not sure it's used much elsewhere in the documentation. Personally I think PATTERN mode is a lot clearer. (As much as I like the nostalgia of TRACKER mode).

PR #121, opened 2017-10-23
Fixes #119 by saturating multiplication.

### LINKED ISSUE #119: Multiply Bug and Saturation
#### Please describe the bug.

Multiply currently zeroes after overflows, goes to INT16_MIN in exact overflows.

The behaviour should probably be to saturate at the boundaries.

```
A 1
L 1 15: A * A 2
> A => -32768
L 1 16: A * A 2
> A => 0
```

burnsauce:
#### What does this PR do?

Causes `MUL` to saturate at int16_t limits.

`* 16384 3`
`* -16384 3`

The existing behaviour was inconsistent and useless.  The new behaviour is to saturate, but it could equally modulo.  I didn't think modulo was appropriate for multiplication, though it may be for other operations.

#### I have,
* [x] updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## vNEXT
 - **NEW**: added a cheat sheet PDF
+- **IMP**: screen now redraws only lines that have changed
 
 ## v2.1
 - **BREAKING**: the `I` variable is now scoped to the `L` loop, and does not exist outside of an execution context.  Scripts using `I` as a general-purpose variable will be broken. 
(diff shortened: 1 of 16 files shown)


burnsauce:
The only other thing that _might_ need saturation is raw numbers.
```
> -32769
32767
> 32768
-32768
```
At least it's consistent here, the number ends up wrapping.  It's an odd case because it's technically user error to enter an out-of-bounds number, but the results might be quizzical.  Saturating this value would be consistent.

PR #123, opened 2017-10-23
Blanking screensaver with 90 min timeout

burnsauce:
#### What does this PR do?

Blanks the Teletype display after 90 minutes of keystroke activity.  Requires a keystroke to wake the display.  That keystroke is dropped.

https://llllllll.co/t/teletype-screen-saver/9769

Set the timeout (see screensaver_mode.h) to 10 seconds, then cycle through modes, use the teletype as normal.  Every 10 seconds after your last keystroke, the display should blank.  Press any key (except for mods SHIFT, ALT, and CTRL) to wake the display.  It should reveal the last screen you were on, although it will be as if you had just navigated to the screen (input and message lines will be blank, etc).

Static patterns on the display will cause bright cells to deteriorate more quickly than black cells.  It's somewhat healable with OLED, but would need a noisy animation to play for a while to do so.

There is another strategy to code this: by shortcutting handler_ScreenRefresh.  It would be slightly more efficient, but it won't allow the screensaver interface to be extensible.  I erred on the side of flexibility.

#### I have,
* [x] updated CHANGELOG.md


burnsauce:
Don't merge this until #122 is merged.  I need to update the code to reflect the new UI protocol.  If it is merged, I will push a new commit to this PR.

burnsauce:
Additionally, I will be adding button press and knob movement as conditions to wake the screensaver.

Issue #129, opened 2017-11-06
JI OP Broken (Just Intonation)


cmcavoy:
#### Please describe the bug.
JI OP is broken according to [post on lines](https://llllllll.co/t/teletype-code-exchange/839/436). Tracking it here.

I haven't confirmed this bug.

cmcavoy:
Confirmed that this is still an issue. I asked for more information from @galapagoose, he was kind enough to send some more background,

> I think I had some pseudo-code lying around that should fix it - only issue is it needed a log() lookup which seemed like it would have to be a custom LUT just for this op. I've resisted digging into the TT codebase so I'm not exactly sure how it would best be implemented.
> 
> If you're wanting to pursue it / post the info, the requirement specs are:
> Input is 2 integers <num> <denom>
> The output is a number in the range 0 to V 1
> 
> Here's some code w/ bits of pseudo-code slotted in. 
> 

```
static void op_JI() {
	uint32_t n = pop();
	uint32_t d = pop();

	// normalize fraction between 1(root) and 2(octave)
	while( n < d ){
		n <<= 1;
	}
	while( n >= (d << 1) ){
		d <<= 1;
	}

	uint32_t ji = ( n << 8 ) / d; // 24bit LUT base

	// split `ji` into LUT index & fractional coefficient for xfade
	uint8_t ix = ji >> 16;
	uint8_t c  = (uint8_t)(( ji >> 8 ) & (uint32_t)0xFF );

	// lookup
	uint32_t a = ji_lut[ix];
	uint32_t b = ji_lut[ix+1]; // needs edge-check

	// xfade (integer math fun...)
	ji = a + c*(b-a)

	push( ji >> 4 ); // shift right to undo x16 range expansion in table
}
```

> And this is the setup details, though this should likely be done offline and stored in flash rather than at startup.


```
#define LUT_SIZE 256

// create the LUT (pre-compute & save in flash)
uint16_t ji_lut[LUT_SIZE];
for( uint8_t i=0; i<LUT_SIZE; i++ ){
	// calculate as float / double then cast to u32
	ji_lut[i] = (uint32_t)( (5441.318221
		                        * log_f(0.12231220585 * (float)(i+LUT_SIZE) / (float)LUT_SIZE)
		                      + 4965.36721229)
		                    * 16.0 // expand dynamic range *16 (still fits in s16)
	                      );
}
```

cmcavoy:
[...]

PR #133, opened 2017-11-17
JI op corrected to map to volt/octave values

trentgill:
#### What does this PR do?
Corrects the functionality of the JI op. It previously gave raw ratio conversions that didn't relate to the 1volt / octave standard in Eurorack.

Uses a prime-factorization technique, which goes up to 13-limit tuning. Should be sufficient for most musical use-cases. Every extra prime extends worst case search time for invalid ratios.

Presently an invalid ratio (one which cannot be factored with primes of
13 or less) simply returns 0 (the base pitch). All results are scaled
between 0 and V 1.

The weird list of constants is calculated with a logarithmic function, representing some kind of musical fraction of the octave. These constants can simply be composed with addition (numerator) and subtraction (denominator) to find the JI result. Prime factorization is used to break down composite input into it's prime components.

I tried to add these tests in process_tests.c but they would always fail. Any ideas how? I can push those tests too if I can figure out why they didn't work
"JI 0 0" -> 0
"JI 3 2" -> 958
"JI 1323 -1024" -> 605
"JI -40 39" -> 60
"JI 23 13" -> 0

https://github.com/monome/teletype/issues/129

#### I have,
* [x] updated CHANGELOG.md
No changes to documentation required.

I also wrote a little script that generates the constants stored in the function. I would normally include them in a block comment immediately above the function, but it didn't seem to fit with the style. Is there somewhere more comments or info are stored about the code?

Here's the function for reference:
int16_t ji_find_prime_constant( uint16_t prime )
{
	float r = 1638.0 * logf( (float)prime ) / log( 2.0 );
	while( r > 1638.0 ){ r -= 1638.0; }
	r *= 16.0;
	return( (int16_t)( r + 0.5 ) );
}

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## vNEXT
 - **NEW**: added a cheat sheet PDF
+- **FIX** : corrected functionality of JI op for 1volt/octave tuning
 
 ## v2.1
 - **BREAKING**: the `I` variable is now scoped to the `L` loop, and does not exist outside of an execution context.  Scripts using `I` as a general-purpose variable will be broken. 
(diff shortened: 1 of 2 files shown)


cmcavoy:
bitmoji

PR #135, opened 2017-12-14
2.2 Features

burnsauce:
#### What does this PR do?

- **NEW**: new bitwise ops: &, |, ^, ~, BSET, BCLR, BGET
- **NEW**: new ops PARAM.SCALE min max and IN.SCALE min max to add static scaling to inputs
- **NEW**: blanking screensaver after 90 minutes of keyboard inactivity, any key to wake
- **NEW**: new op: CHAOS chaotic sequence generator.  Control with CHAOS.ALG and CHAOS.R
- **NEW**: new op family: INIT to clear state
- **NEW**: new ops: R, R.MIN, and R.MAX
- **IMP**: optional profiling (enable at build time)
- **IMP**: screen now redraws only lines that have changed
- **FIX**: multiply now saturates at limits, previous behaviour returned 0 at overflow
- **FIX**: entered values now saturate at int16 limits
- **FIX**: reduced flash memory consumption by not storing TEMP script

https://llllllll.co/t/2-2-release-status-beta-1/9778

Extensively :) 

#### I have,
* [x] updated CHANGELOG.md


scanner-darkly:
the only thing that i would maybe add is a comment in `flash.c` around calculating the size to explain there is a temp scene. please note i did not review chaos/init implementation as i wasn't following that discussion closely.

src/teletype.h
```diff
@@ -8,8 +8,8 @@
 #include "command.h"
 #include "state.h"
 
-#define TELETYPE_VERSION "TELETYPE 2.1.0"
 #define TELE_ERROR_MSG_LENGTH 16
+#define TELETYPE_PROFILE // un-comment this line to enable profiling
```
tehn: profiling is *on* then, right?

should probably turn this off for release.

burnsauce: Oops!  Yes, comment that out for release.

tehn:
i'm going to merge this.

i see that *some* of the docs are in there already, partial stubs at least. let's get the remainder merged separately before i make a public release so the docs are current.

tehn:
@burnsauce would you like the honor of closing any relevant git issues that this addressed?

burnsauce:
Only one of them (CHAOS) was not marked with FIX in the commits, so I closed it manually.

PR #164, opened 2018-10-12
DEL.X / DEL.R ops

alpha-cactus:
#### What does this PR do?
added ```DEL.X``` and ```DEL.R``` ops allowing for chained(multiple) delays at spaced intervals.  added common delay add function for all delay ops to save code space.  additionally increased the maximum number of delays from 8 to 16.

https://llllllll.co/t/teletype-3-feature-requests-and-discussion/16219/27?u=alphacactus

tested both old and new delay ops with variable delay times and number of delays.  tested edge cases such as calling ```DEL.X``` for zero delays, delay times of zero, and negative delay times.  

#### I have,
* [x] updated CHANGELOG.md

### DIFF
--- a/src/ops/delay.c
+++ b/src/ops/delay.c
@@ -4,9 +4,9 @@
 #include "teletype.h"
 #include "teletype_io.h"
 
-static void delay_common_add(scene_state_t *ss, exec_state_t *es, 
-                        int16_t i, int16_t delay_time,
-                        const tele_command_t *post_command);
+static void delay_common_add(scene_state_t *ss, exec_state_t *es, int16_t i,
+                             int16_t delay_time,
+                             const tele_command_t *post_command);
 
 static void mod_DEL_func(scene_state_t *ss, exec_state_t *es,
                          command_state_t *cs,
@@ -16,28 +16,30 @@ static void op_DEL_CLR_get(const void *data, scene_state_t *ss,
                            exec_state_t *es, command_state_t *cs);
 
 static void mod_DEL_X_func(scene_state_t *ss, exec_state_t *es,
-                         command_state_t *cs,
-                         const tele_command_t *post_command);
+                           command_state_t *cs,
+                           const tele_command_t *post_command);
 
 static void mod_DEL_R_func(scene_state_t *ss, exec_state_t *es,
-                         command_state_t *cs,
-                         const tele_command_t *post_command);
+                           command_state_t *cs,
+                           const tele_command_t *post_command);
 
 const tele_mod_t mod_DEL = MAKE_MOD(DEL, mod_DEL_func, 1);
 const tele_op_t op_DEL_CLR = MAKE_GET_OP(DEL.CLR, op_DEL_CLR_get, 0, false);
 const tele_mod_t mod_DEL_X = MAKE_MOD(DEL.X, mod_DEL_X_func, 2);
 const tele_mod_t mod_DEL_R = MAKE_MOD(DEL.R, mod_DEL_R_func, 2);
 
-//common code to queue a delay shared between all delay ops
-//NOTE it is the responsibility of the callee to call tele_has_delays
-static void delay_common_add(scene_state_t *ss, exec_state_t *es, 
-                        int16_t i, int16_t delay_time,
-                        const tele_command_t *post_command) {
+// common code to queue a delay shared between all delay ops
+// NOTE it is the responsibility of the callee to call tele_has_delays
+static void delay_common_add(scene_state_t *ss, exec_state_t *es, int16_t i,
+                             int16_t delay_time,
+                             const tele_command_t *post_command) {
+    if (i < DELAY_SIZE) {
         ss->delay.count++;
         ss->delay.time[i] = delay_time;
         ss->delay.origin_script[i] = es_variables(es)->script_number;
         ss->delay.origin_i[i] = es_variables(es)->i;
         copy_command(&ss->delay.commands[i], post_command);
+    }
 }
 
 static void mod_DEL_func(scene_state_t *ss, exec_state_t *es,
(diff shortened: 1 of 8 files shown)


src/ops/delay.c
```diff
@@ -4,35 +4,56 @@
...
 const tele_op_t op_DEL_CLR = MAKE_GET_OP(DEL.CLR, op_DEL_CLR_get, 0, false);
+const tele_mod_t mod_DEL_X = MAKE_MOD(DEL.X, mod_DEL_X_func, 2);
+const tele_mod_t mod_DEL_R = MAKE_MOD(DEL.R, mod_DEL_R_func, 2);
+
+//common code to queue a delay shared between all delay ops
+//NOTE it is the responsibility of the callee to call tele_has_delays
+static void delay_common_add(scene_state_t *ss, exec_state_t *es, 
+                        int16_t i, int16_t delay_time,
+                        const tele_command_t *post_command) {
+        ss->delay.count++;
```
scanner-darkly: i would add a check here to make sure it doesn't exceed `DELAY_SIZE` - imho it's a good idea to have safety checks in helper functions.

scanner-darkly:
other than the 1 comment looks good! 
did you run clang-format as well? (we should add it to the checklist)

src/ops/delay.c
alpha-cactus: ah good catch I can add that in.  on the same topic do you think it would be good to move ```tele_has_delays``` into the helper function as well?  the downside is it would get excess times times by ```DEL.X```, but would make the helper function slightly more robust for other future uses.

alpha-cactus:
no I didn't run clang-format.  never used it before could you go over briefly how to do that properly?

src/ops/delay.c
scanner-darkly: i'd leave it out - the function has a well defined and logical contract (it adds a delay), `tele_has_delays` is outside of that (and only has meaning to where this function is called from).

scanner-darkly:
re: clang - check this: https://github.com/monome/teletype#code-formatting

you'll need to install clang-format first. if you're on windows you might also need to modify the make file (or you could prob run it from bash)

alpha-cactus:
I ran clang-format using ```make format``` and it changed more than the files I was working on.  the changes from what I tell seem okay.  I'm going to build and test in a minute here. 

[clang_diff.txt](https://github.com/monome/teletype/files/2474964/clang_diff.txt)

scanner-darkly:
@alpha-cactus that’s weird... i ran it before i did a pull request for 3.0, so there shouldn’t be any major changes. make sure it’s using the clang config included in the repo. if it does just leave it out for now and i’ll investigate - entirely possible i used the wrong config..

alpha-cactus:
it is as far as I can tell.  I'm on OSX,so I installed clang-format using ```brew install clang-format``` if that might be the cause.

screenshot of running ```make format```
screen shot 2018-10-12 at 10 27 48 pm

scanner-darkly:
thanks for confirming! i'll take a look then, if i did it wrong on my side i'll need to fix several files. we can just skip the step for this PR.

src/ops/delay.c
```diff
@@ -42,3 +65,67 @@ static void op_DEL_CLR_get(const void *NOTUSED(data), scene_state_t *ss,
...
+        // queue the delay
+        delay_common_add(ss, es, i, delay_time_next, post_command);
+
+        // increment delay time for next delay
+        // normalise incremented value to stop negative wrap from increment
+        delay_time_next += delay_time;
+        delay_time_next = normalise_value(1, 32767, 1, delay_time_next);
+
+        num_delays--;
+        i++;
```
scanner-darkly: sorry, should've looked into the code closer the 1st time.. there is a problem with this spot here. for multiple delays it will find the 1st available slot but for any consecutive delays it will use the next delay without checking if it's free or not. you have to check that for each delay added. 

this: `while (ss->delay.time[i] != 0 && i != DELAY_SIZE) i++;` can be moved to `delay_common_add` and `i` parameter can be removed, so the function's responsibility will be finding the 1st available spot and adding a delay there. it could return a bool indicating if it was added, so if `false` is returned this loop can assume no more spots are available and exit early.

alpha-cactus: wow another good catch.  elegant solution too.  I'll work on getting that together tonight.

src/ops/delay.c
```diff
@@ -30,34 +30,37 @@ const tele_mod_t mod_DEL_R = MAKE_MOD(DEL.R, mod_DEL_R_func, 2);
...
+static bool delay_common_add(scene_state_t *ss, exec_state_t *es,
                              int16_t delay_time,
                              const tele_command_t *post_command) {
+    int16_t i = 0;
+
+    // 0 is the magic number for an empty slot.
+    // Be careful not to set delay.time[i] to 0 before calling this function.
+    while (ss->delay.time[i] != 0 && i != DELAY_SIZE) i++;
+
+    if (delay_time < 1) delay_time = 1;
```
alpha-cactus: added a check for the delay time to the common code function. removed the check from the ```DEL``` op function, but it's still in the ```DEL.X``` and ```DEL.R``` because the wrap check would cause 32ms delays if the delay time param was negative.

src/ops/delay.c
```diff
@@ -69,30 +72,23 @@ static void op_DEL_CLR_get(const void *NOTUSED(data), scene_state_t *ss,
...
-    while (ss->delay.time[i] != 0 && i != DELAY_SIZE) i++;
+    if (delay_time < 1) delay_time = 1;
 
-    while (i < DELAY_SIZE && num_delays > 0) {
-        // queue the delay
-        delay_common_add(ss, es, i, delay_time_next, post_command);
+    // set first delay time to delay time
+    delay_time_next = delay_time;  
 
+    while ( num_delays > 0 && delay_common_add(ss, es, delay_time_next, post_command) )
```
alpha-cactus: the function call is part of the while check instead of using a separate bool variable.  pretty sure this is fine, but I may be missing something.

PR #165, opened 2018-10-17
Switch to using `git-clang-format` for `make format`

samdoshi:
#### What does this PR do?

This PR updates the `make format` target use `git-clang-format`, which has been set to format _only_ staged and unstaged changes, instead of formatting all the code in the repo.

This should avoid issues like in #164 where new users end up formatting lots of old code, and also stop the `blame` view getting messed up. However it does mean that `make format` needs to be run each commit (which it should be anyway).

There is a `make format-all` command that has the old behaviour, I expect we'll need to run this occasional to pick up any missing changes.

I'm not 100% certain this is better than how we do it currently, but I think it is. Thoughts?

Run `make format` and `make format-all`

#### I have,
* [ ] updated CHANGELOG.md

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,4 +32,7 @@ teletype.zip: clean-zip
 		docs/cheatsheet/cheatsheet.pdf
 
 format:
+	git-clang-format -f --style=file
+
+format-all:
 	find . -type f -name "*.c" -o -name "*.h" | xargs clang-format -style=file -i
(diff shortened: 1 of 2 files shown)


samdoshi:
Added a note about `make format` to `PULL_REQUEST_TEMPLATE.md` (screwed up the change a few times so had to force push...)

I'll merge in a few days assuming no objections

samdoshi:
FYI I'm pretty sure I can modify the `.travis.yml` file to fail if there are any differences between the repo and `make format-all`.

Any interest?

scanner-darkly:
sounds like a good idea!

PR #170, opened 2018-12-02
require confirm for flash initialization

tehn:
#### What does this PR do?

on flash_reset detection (checking magic number in `flash_prepare`) blink a light and wait for confirmation by pressing key down. otherwise it times out and doesn't run.

works perfectly well for newly flashed units. we'd need to see the bug-state to trigger this, which isn't known. it's a protection measure at this point.

#### I have,
* [ ] updated `CHANGELOG.md`

### DIFF
--- a/module/flash.c
+++ b/module/flash.c
@@ -4,6 +4,7 @@
 
 // asf
 #include "flashc.h"
+#include "gpio.h"
 #include "init_teletype.h"
 #include "print_funcs.h"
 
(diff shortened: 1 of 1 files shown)


tehn:
## note

this is a _confusing_ mechanism, to protect flash.

if you're flashing "fresh" from git you _must_ hit the button within the flashing-period, otherwise boot will fail (haven't had a minute to understand why)

if you're flashing from a hex file with presets the magic number will already be set and you won't be presented with the flashing-need-for-pressing-button

PR #171, opened 2018-12-03
fixes for flash initialization

scanner-darkly:
#### What does this PR do?

further fixes for flash initialization:
- added a screen with a warning
- previous change would always init with blank scenes
- added a safeguard for preset_no

#### I have,
* [ ] updated `CHANGELOG.md`


scanner-darkly:
also fixed some formatting, when doing a compare use `w=1` in the URL to ignore whitespace changes.

PR #178, opened 2019-06-14
New ansible remote ops (`ANS.G` etc)

csboling:
#### What does this PR do?

Adds several new Ansible remote ops:
* `KR.CUE` allows cueing a Kria pattern.
* `KR.PG` allows getting/setting a numerical code corresponding to the active Kria page.
* `ANS.APP` allows getting/setting a numerical code corresponding to the active Ansible app.
* `ANS.G` allows simulating a grid event on Ansible, e.g. `ANS.G 0 0 1` to simulate the top left key of the grid connected to Ansible being held down. `ANS.G.P x y` simulates a press, so an on event immediately followed by an off event. `ANS.G.LED x y` allows reading the current LED brightness stored in Ansible's `monomeLedBuffer`. `ANS.A` and `ANS.A.LED` are analogous for arc.

https://llllllll.co/t/teletype-3-feature-requests-and-discussion/16219/291

Connect an Ansible to Teletype's ii bus, and perhaps a grid to Ansible. Use `ANS.APP` to get/change the active app.  Use `ANS.G` and `ANS.G.P` to simulate grid key events, observe Ansible modify the state of the connected grid. Use `ANS.G.LED` to read the LED brightness at a given position. Do the same for `ANS.A` and `ANS.A.LED` ops with arc and its apps.

#### I have,
* [x] updated `CHANGELOG.md`
* [ ] run `make format` on each commit - I ran this and it changed a bunch of spots/files that this PR does not touch, so I'm just checking in what it did to ansible.c. Doing a full pass with this seems like it should maybe be its own PR.


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "libavr32"]
 	path = libavr32
-	url = https://github.com/monome/libavr32.git
+	url = https://github.com/csboling/libavr32.git
```
tehn: this seems like an error?

tehn:
looks great otherwise!

tehn:
looking good! would you like this merged?

csboling:
Yeah that would be great. Just looked at it again and found those stray debug prints. Along with SCRIPT.POL I think this is everything I have for Teletype 3.1.

PR #179, opened 2019-07-04
bugfix: git sha1 script requires bash

csboling:
#### What does this PR do?

Without specifying a shell, make will use `/bin/sh`, which on some environments (Ubuntu, and therefore [this](https://hub.docker.com/r/dewb/monome-build/) Dockerfile) is not `bash`. This causes redirection to fail in the bit that generates the git sha1 to embed in the firmware. Also `git describe --always --dirty` was winding up with two different version numbers in the string displayed on Teletype, like `3.0.0 v2.2.0-alpha.7-244` as well as the wrong tag for some reason, this change adds `--exclude '*'` to omit tags in this part so you just get a version + sha like `3.0.0 0F3DAA7-DIRTY`.

This was causing some confusion because the version number was coming out weird from my builds: [1](https://llllllll.co/t/teletype-3-feature-requests-and-discussion/16219/348), [2](https://llllllll.co/t/teletype-3-0/11232/421)

Build ye Firm-Ware, flafhe ye Module.

#### I have,
* [x] updated `CHANGELOG.md` (N/A)
* [x] updated the documentation (N/A)
* [x] run `make format` on each commit (N/A)

### DIFF
--- a/module/Makefile
+++ b/module/Makefile
@@ -60,6 +60,8 @@
 # \asf_license_stop
 #
 
+SHELL=bash
+
 # Makefile.avr32.in defines an unused variable build, which is used in the clean
 # target, it's probably there to list other build targets
 build += ../src/scanner.c ../src/match_token.c ../module/gitversion.c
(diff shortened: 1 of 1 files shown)


tehn:
thank you!

PR #198, opened 2020-04-13
FADER.SCALE, FADER.CAL.MIN/MAX, FADER.CAL.RESET

discohead:
#### What does this PR do?

Adds scaling for `FADER` and `FB` just like there is for `IN` and `PARAM`

Admittedly feels a bit naughty putting ii device related functionality along side the onboard hardware stuff in `main.c` and `src/state`, but I wasn't sure what might be a better approach. Open to suggestions.

A 16n is required. I have the Sweet Sixteen, so someone with the original should test it.

The scaling and calibration behavior should be exactly the same as it is for `IN` and `PARAM`.

#### I have,
* [x] updated `CHANGELOG.md`


tehn:
i don't have a 16n, so any additional testing would be great.

getting ready to the next TT version release, so this could be added

discohead:
@tehn it works perfectly with the Sweet Sixteen. I don't have access to an original 16n, it should work fine though. maybe @scanner-darkly ?

scanner-darkly:
i'll give it a test tonight or tomorrow!

scanner-darkly:
sorry, it's been a hectic couple of days, will get to it this afternoon!

scanner-darkly:
sorry - looks like i won't be able to test. my 16n currently has the leader firmware, and i'm having difficulty updating the firmware.

in any case, i don't think testing on a different 16n is really necessary - the code isn't doing anything that would be hardware specific, it just reads values the same way `FADER` op does.

PR #217, opened 2020-07-27
improve performance when navigating in docs

midouest:
#### What does this PR do?

Previously, the `onchangehash` handler would run `fixBr` every time
the user navigated to a new hash. The `fixBr` function takes a long
time to execute because it causes lots of reflows. This commit
removes the (seemingly) unnecessary `linksToNewTab` and `fixBr` function calls
from the onhashchange handler by moving the necessary code to a new helper
function.

1. Build docs
2. Open `teletype.html`
3. Navigate to various sections in the navigation pane (should be almost instant now)
4. Try navigating backward/forward in the browser

#### I have,
* [X] updated `CHANGELOG.md`

### DIFF
--- a/utils/templates/template.html
+++ b/utils/templates/template.html
@@ -73,13 +73,13 @@
         <svg style="display: none;">
             <symbol id="custom-hamburger-icon" viewBox="0 0 22 17"
                 xmlns="http://www.w3.org/2000/svg">
-                    <g 
+                    <g
                         id="Page-1"
                         stroke-width="1"
                         fill="none"
                         fill-rule="evenodd"
                         stroke-linecap="round">
-                            <g 
+                            <g
                                 id="Artboard"
                                 transform="translate(-180.000000, -130.000000)"
                                 stroke-width="2">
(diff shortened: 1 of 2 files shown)


utils/templates/template.html
```diff
@@ -209,17 +209,17 @@ <h1 class="title">$title$</h1>
...
                             .classList
-                            .contains("NavigationArea-navLi--secondary") && 
+                            .contains("NavigationArea-navLi--secondary") &&
                         navElements[j]
                             .getAttribute("active-parent-anchor") !== hrefClicked) {
                                 navElements[j]
                                     .classList
                                     .add("NavigationArea-navLi--closed");
-                    } else if (parentHrefClicked && 
+                    } else if (parentHrefClicked &&
```
midouest, line 222: Above changes were auto-formatting from my IDE

scanner-darkly:
can't comment on the changes (not my area of expertise) so just wanted to say - thank you for fixing this, it pretty much made the manual unusable on ipad.

csboling:
Thanks! I routinely forget about this and try to pull up TT docs on mobile only to get frustrated and go to my computer.

This won't be reflected in the monome.org docs automatically, the HTML manual has to get built and the generated file(s) committed to the monome/docs repo. @midouest if you wanted to PR the updated documentation to docs it would be much appreciated!

midouest:
@csboling Done! https://github.com/monome/docs/pull/263

scanner-darkly:
just tried it - much faster! 👍

PR #225, opened 2020-09-07
Add QT.S, QT.CS, QT.B, VN, BTOG, DEL.B

desolationjones:
#### What does this PR do?
- **NEW**: bitmasked delay and quantize: `DEL.B..`,  `QT.B..`
- **NEW**: scale and chord quantize: `QT.S..`, `QT.CS..`
- **NEW**: bit toggle OP: `BTOG..`
- **NEW**: volts to semitones helper OP: `VN..`
- **IMP**: DELAY_SIZE increased to 64 from 16
- **FIX**: scale degree arguments 1-indexed: `N.S`, `N.CS`
- **NEW**: `.gitattributes` file to set LF as default EoL char to prevent automatic use of CRLF on GitHub Desktop on Windows

https://llllllll.co/t/teletype-3-2-feature-requests-and-discussions/32052/

`VN N x` should always return `x`.
`QT.S`, `QT.CS`, and `QT.B` OPs can be tested with note number inputs (for both input and root args) between -127 to 127 and the outputs should adhere to the scales selected. It is easiest to use an `L` loop and write the inputs and outputs to two columns of patterns to compare side-by-side.
`BTOG x y` can be tested on the Live screen by making sure that repeated toggles of bit `y` alternately adds and subtracts `2 ^ (y - 1)` from `x`.
`DEL.B` could be tested by something like:
Script 1 (manually trigger):
`$ 2`
`INIT.P.ALL`
`P.I 0`
`DEL.B X B: $ 2`
Script 2:
`P.HERE LAST SCRIPT`
`P.I + P.I 1`
The time values in the pattern should reflect the active intervals (1 bits of mask `B`) of time `X`.

These OPs mostly round out the behavior of bitwise manipulation synergizing with new `B` number entry. `BTOG` is a needed helper for generative bitbeat(?) scenes.
The quantize OPs fix up a lot of the edge cases inherent to manually-scripted quantize scenes while also greatly compressing that cliche to a single line. The use of note numbers rather than CV values was a sticking point for me but polled users unanimously prefer the note numbers. `VN` becomes a useful helper here because it is not *exactly* equivalent to `/ X N 1`; it handles more like a quantizer, which solves the "rounding gap" when `X` passes from negative to positive.

N/A

#### I have,
* [x] updated `CHANGELOG.md`
* [ ] run `make format` on each commit (I can't seem to `make format` in docker because it doesn't include `clang-format`)


tehn (approved):
looks great! @scanner-darkly any feelings about a point release soon?

scanner-darkly (approved):

scanner-darkly:
some indentation is off, i can run clang-format and submit a PR.

@tehn - i think new w/ ops is the only outstanding thing, i'll comment in that PR!

PR #227, opened 2020-10-11
Adds title frontmatter to docs for HMTL metadata

nattog:
Adds title metadata for generated HTML documentation. Deletes HTML chunk which displays metadata on the page, hidden behind the menu, and not rendered when the title is not set

Tested it working on a local build
I already made a PR here https://github.com/monome/docs/pull/271 before learning they were generated.

### DIFF
--- a/docs/intro.md
+++ b/docs/intro.md
@@ -1,3 +1,7 @@
+---
+title: teletype - manual
+---
+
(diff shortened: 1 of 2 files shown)


tehn:
Thank you!

PR #233, opened 2021-03-13
preserve number format

scanner-darkly:
#### What does this PR do?

previously, numbers entered using `X..` | `B..` | `R..` formats would get converted to decimal whenever a line containing them would be entered. with this change the number format will be preserved.

ensure number formats are preserved when:
- entering them in live screen and then using history
- entering numbers in script editing
- storing scenes to USB and restoring them from USB

test `X0`, `B0`, `R0`.

#### I have,
* [ ] updated `CHANGELOG.md` - no need, it restores expected functionality
* [ ] updated the documentation - no need, it restores expected functionality

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -82,19 +82,19 @@ void print_command(const tele_command_t *cmd, char *out) {
                 break;
             }
             case XNUMBER: {
-                char number[5];
+                char number[6];
                 itoa_hex(value, number);
                 strcat(out, number);
                 break;
             }
             case BNUMBER: {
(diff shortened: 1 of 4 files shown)


attejensen:
This looks great! Didn't test to/from USB, but the rest is working as expected.

scanner-darkly:
apologies for multiple commits - this should be good for a review now.

tehn (approved):

tehn:
this is a really nice addition, thank you!

PR #234, opened 2021-03-14
fixed R... + reversed N.B and N.BX

attejensen:
#### What does this PR do?
Fix `R...` and change the bit-order of OPs N.B and N.BX, by request the number of scales is bumped from 8 to 16

https://llllllll.co/t/teletype-3-2-feature-requests-and-discussions/32052/343

`B100 = R001`
`B110 = R011`

`N.B 0 R101011010101` should setup C-major-scale
`N.B 0 -1` should setup C-dorian

N/A

N/A

#### I have,
* [ ] updated `CHANGELOG.md`


docs/ops/maths.toml
```diff
@@ -287,7 +287,7 @@ prototype = "N.B d"
 prototype_set = "N.B r s"
 short = "get degree `d` of scale/set scale root to `r`, scale to `s`, `s` is either bit mask (`s` >= 1) or scale preset (`s` < 1)"
 description = """
-Converts a degree in a user-defined equal temperament scale to a value usable by the CV outputs. Default values of `r` and  `s` are 0 and B101011010101, corresponding to C-major.
+Converts a degree in a user-defined equal temperament scale to a value usable by the CV outputs. Default values of `r` and  `s` are 0 and R101011010101, corresponding to C-major.
```
scanner-darkly, line 290: would it be helpful to add a note explaining why `R` notation here makes more sense?

scanner-darkly: also should probably mention this op will only use 12 lower bits?

src/command.c
```diff
@@ -1,6 +1,7 @@
 #include "command.h"
 
 #include <string.h>  // memcpy
+#include "helpers.h"
```
scanner-darkly, line 4: is it actually used here?

src/command.c
```diff
@@ -57,14 +58,19 @@ static void itoa_rbin(uint16_t value, char *out) {
...
-            out[index++] = '0' + v;
-            dont_ignore_zeros = 1;
-        }
+        out[index++] = '0' + v;
     }
-
-    if (!dont_ignore_zeros) out[index++] = '0';
     out[index] = '\0';
+
+    if (!dont_ignore_zeros) {
```
scanner-darkly, line 65: you don't need this check as `dont_ignore_zeros` is never changed above.

src/command.c
```diff
@@ -57,14 +58,19 @@ static void itoa_rbin(uint16_t value, char *out) {
...
+        for (int8_t i = 16; i > 0; i--) {
+            if (out[i] == '1') {
+                index = i;
+                break;
+            }
+        }
+        out[index + 1] = '\0';
```
scanner-darkly, line 72: if i'm reading correctly, with the value of `0` it will incorrectly output `0000000000000000` and it will place an extra `\0` out of bounds of `out` array.

src/helpers.c
```diff
@@ -94,12 +93,26 @@ const char *to_voltage(int16_t i) {
...
+
+long int tokentol_reverse(const char* str, char** endptr, int base) {
+    long int reverse_val;
+    reverse_val = strtol(strrev(str), endptr, base);
+    return reverse_val;
+}
+
+const char* strrev(const char* str) {
+    int index = strlen(str);
+    char* revStr = (char*)malloc(index--);
```
scanner-darkly, line 113: this function will create a memory leak as it allocates memory each time it's called

scanner-darkly:
added some comments!
also, could you add the new ops to the built in help screen? it's in `module/help_mode.c` file.

src/command.c
attejensen: Removed dont_ignore_zeros entirely

src/command.c
attejensen: Opps, old left over, removed

docs/ops/maths.toml
attejensen: Done

module/help_mode.c
```diff
@@ -327,6 +327,11 @@ const char* help5[HELP5_LENGTH] = { "5/16 OPERATORS",
 									"     AND EACH SUBSEQUENT BIT",
 									"     IS +1 SEMITONE",
                                     " ",
+                                    "// N.B  D   GET DEGREE IN SCALE"
```
scanner-darkly, line 330: don't need `//` before the ops, it's used to separate sections

src/command.c
```diff
@@ -52,25 +50,28 @@ static void itoa_bin(uint16_t value, char *out) {
...
-        for (int8_t i = 16; i > 0; i--) {
-            if (out[i] == '1') {
-                index = i;
-                break;
-            }
-        }
-        out[index + 1] = '\0';
+    index = 0; // catch this below if all 0s
+    for (int8_t i = 16; i > 0; i--) {
+	if (out[i] == '1') {
```
scanner-darkly, line 65: indent seems off

src/command.c
```diff
@@ -52,25 +50,28 @@ static void itoa_bin(uint16_t value, char *out) {
...
+	    index = i;
+	    break;
+	}
+    }
+    if (index == 0){     // all 0s....
```
scanner-darkly, line 70: you don't need this whole `if` block if you simply set `index` to 1 above

src/helpers.c
```diff
@@ -102,17 +101,20 @@ int16_t bit_reverse(int16_t unreversed, int8_t bits_to_reverse) {
...
-    while (0 <= index) revStr[destIndex++] = str[index--];
-    revStr[destIndex] = 0;
-    return revStr;
+const char* strrev(const char* str){
+    int8_t length = strlen(str);
+    char rev[length];
+    for(int8_t i = 0; i<length; i++){
+	rev[length-1-i] = str[i];
+    }
+    strcpy(str,rev);
```
scanner-darkly, line 117: i think you need to include `\0` at the end of `rev` in order for `strcpy` to work properly. also, i'd make it return `void` since you're reverting the string passed as a parameter (and this also means it shouldn't be marked as `const` parameter).

i would just keep it simple, it's only used in one place and we know it's for a 16 char string, so you could just revert in place:

```
char c;
for (int8_t i = 0; i < 8; i++) {
    c = str[i];
    str[i] = str[15 - i];
    str[15 - i] = c;
}
```

module/help_mode.c
attejensen: Ok, thanks

src/command.c
attejensen: odd from trusty editor...

src/command.c
attejensen: it might have been updated in the for loop just above in case a 1 ws found (line 71) ...

src/helpers.c
attejensen: this was getting too convoluted anyhow, replaced everything with a single, simple rev_bitstring_to_int in helpers

src/command.c
attejensen: ...but there were other mistakes in there, should be fixed now...

src/command.c
attejensen: shouldn't tabs/space indent and such be caught by make format?

src/helpers.c
```diff
@@ -94,12 +91,20 @@ const char *to_voltage(int16_t i) {
...
+            reversed |= 1 << ((bits_to_reverse - 1) - i);
     }
     return reversed;
 }
+
+int16_t rev_bitstring_to_int(const char* token) {
+    int8_t length = strlen(token);
+    int16_t value = 0;
+    for (int8_t i = 0; i < length; i++) {
+        if (token[i] == '1') { value += 1 << i; }
```
scanner-darkly, line 107: i'm not sure this will work properly for negative values, i would use `value |=` instead of `value +=` to be on the safe side.

src/command.c
scanner-darkly: i would expect so. perhaps there is an issue with `make format` - i don't use it as i'm on windows, so i usually run clang format manually.

src/helpers.c
scanner-darkly: i did a quick test and it appears it works properly either way. i would still test with various negative values just to be sure, i used visual studio c++ to test, so this might behave differently.

scanner-darkly (approved):

src/helpers.c
attejensen: Unless I'm mistaken, the | vs + doesn't really matter, since we're just picking up bits set in a string comprised of zeros and ones at this point, thus always generating a positive number here. It works because the tt later chooses to interpret large numbers as negative. I tried manually wrapping value after INT16_MAX (like the tt works) just after this just before return, and it makes no difference, it works as expected either way. In all my tests on the tt it behaves as expected... I don't see any reason why this should be changed, but I'm eager to be proven wrong!

scanner-darkly: but it's not generating positive numbers only. we are adding to a signed 16 bit integer, so that means that once it gets to `1 << 15` that will be interpreted as `-32768` (due to the implicit casting, since it's adding to a signed integer). i think what happens is that due to how two's complement numbers work it ends up giving the correct result either way - with `+=` or `|=` (which might test confirmed, but i tested with c++ instead of c and using a different compiler, so the behaviour might be different). in any case, as long as it's providing the correct result it doesn't matter which way it's done.

attejensen: Agreed!

scanner-darkly:
@tehn i think we're good to merge this one

PR #235, opened 2021-03-15
W/2.0 cleanup

kfirmanty:
#### What does this PR do?
This PR will finish the work related to W/2.0 integration, by adding documentation, possibility to retrieve values from W/2.0 and making sure that all values are send/received properly

https://llllllll.co/t/teletype-and-w-2-0-integration/34220

By executing commands through i2c and verifying that both:
- W/2.0 leds in tape/delay/synth settings change accordingly to value sent
- same value as sent through i2c is received when reading data from W/2.0

Hi,
Sorry for long silence, I tried to fix some outstanding issues previous weekend and will try to put more work into it next weekend.
I have updated the documentation using descriptions from lua config files in crow repo for W/2.0 i2c.
I have also added possibility to retrieve values from W/2.0. The only problem is that this obviously won't work with i2c messages which receive two vals from W/2.0. There is also some messages that expect 2 args when executed from teletype but return 1 val W/2.0 (like FM_RATIO in synth mode https://github.com/monome/crow/blob/main/lua/ii/wsyn.lua#L100) and this obviously won't work with MAKE_GET_SET_OP as it expects for setter to have only one argument. 
If there is anything missing that I not mentioned here please let me know and again thank you for your patience.
#### I have,
* [ ] updated `CHANGELOG.md`

### DIFF
--- a/docs/ops/wslashdelay.toml
+++ b/docs/ops/wslashdelay.toml
@@ -1,0 +1,63 @@
+["W/D.FBK"]
+prototype = "W/D.FBK level"
+short = "amount of feedback from read head to write head (s16V)"
+
+["W/D.MIX"]
+prototype = "W/D.MIX fade"
+short = "fade from dry to delayed signal"
+
+["W/D.LPASS"]
+prototype = "W/D.LPASS cutoff"
+short = "centre frequency of filter in feedback loop (s16V)"
+
+["W/D.FREEZE"]
+prototype = "W/D.FREEZE is_active"
+short = "deactivate record head to freeze the current buffer (s8)"
+
+["W/D.TIME"]
+prototype = "W/D.TIME seconds"
+short = "set delay buffer length in `seconds` (s16V), when rate == 1"
+
+["W/D.LNG"]
+prototype = "W/D.LNG count divisions"
+short = "set buffer loop size as a fraction of buffer time (u8)"
+
+["W/D.POS"]
+prototype = "W/D.POS count divisions"
+short = "set loop location as a fraction of buffer time (u8)"
+
+["W/D.CUT"]
+prototype = "W/D.CUT count divisions"
+short = "jump to loop location as a fraction of buffer time (u8)"
+
+["W/D.FREQ.RNG"]
+prototype = "W/D.FREQ.RNG freq_range"
+short = "TBD (s8)"
+
+["W/D.RATE"]
+prototype = "W/D.RATE multiplier"
+short = "direct `multiplier` (s16V) of tape speed"
+
+["W/D.FREQ"]
+prototype = "W/D.FREQ volts"
+short = "manipulate tape speed with musical values (s16V)"
+
+["W/D.CLK"]
+prototype = "W/D.CLK"
+short = "receive clock pulse for synchronization"
+
+["W/D.CLK.RATIO"]
+prototype = "W/D.CLK.RATIO mul div"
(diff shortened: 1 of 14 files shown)


desolationjones:
For OPs which patch jacks to params, can the docs indicate which param?

kfirmanty:
I will be happy to add them, but right now I was basically copy pasting docs from crow integration:
https://github.com/monome/crow/blob/main/lua/ii/wsyn.lua#L85
and unfortunately I couldn't find the list of available params mapping.

desolationjones:
@trentgill what is the mapping of params for this command? Also, does w/ react the same to a virtual patching/unpatching as it does to a physical cable? e.g. will virtual patching an empty jack still treat it as if a cable is inserted

desolationjones:
This seems ready to merge. I don't know why it has lingered so long, but if there is any area of *specific* concern for testing/feedback please let me know!

docs/ops/wslashdelay.toml
```diff
@@ -0,0 +1,63 @@
+["W/D.FBK"]
+prototype = "W/D.FBK level"
+short = "amount of feedback from read head to write head (s16V)"
+
+["W/D.MIX"]
+prototype = "W/D.MIX fade"
+short = "fade from dry to delayed signal"
+
+["W/D.LPASS"]
```
trentgill, line 9: Could `LPASS` be changed to `FILTER` or `FILT` so it aligns with crow? The specific reason is I've got plans to make the filter more flexible than lowpass only in the future. Using `FILTER` is thus open to changes in the future.

docs/ops/wslashsynth.toml
```diff
@@ -0,0 +1,55 @@
...
+prototype = "W/S.FM.RATIO num den"
+short = "ratio of the FM modulator to carrier as a ratio. floating point values up to 20.0 supported (s16V)"
+
+["W/S.FM.ENV"]
+prototype = "W/S.FM.ENV amount"
+short = "amount of vactrol envelope applied to fm index, -5 to +5 (s16V)"
+
+["W/S.PATCH"]  
+prototype = "W/S.PATCH jack param"
+short = "patch a hardware `jack` (s8) to a `param` (s8) destination"
```
trentgill, line 51: options are:
1 = Ramp
2 = Curve
3 = FM env amount
4 = FM index
5 = Env time
6 = Env symmetry
7 = Gate (opens/closes envelope)
8 = Pitch (v8)
9 = FM numerator
10 = FM denominator
11 = disable

When this command is called, it just changes the mapping, which will have no effect if a cable is not inserted. However, when you do insert a cable, it will use any setting made by PATCH. If you are currently using a different mode, it will successfully 'exit' the current mapping, and switch to the new one.

trentgill (approved):
Approving the PR. Comments inline regarding answers about the `PATCH` command (PATCH affects what will happen when a jack is inserted, it doesn't change *whether* a jack is 'inserted').

The Lowpass -> Filter change would be nice. It's only a couple lines to change. For reference, I initially was using 'lowpass' but changed to 'filter' to keep it open in the future.

Endless thanks for making this happen!

kfirmanty:
Hi just giving heads up, that the work is not abandoned and I will fix conflicts and change the naming this weekend as this and previous weeks were rather frantic on my side.

desolationjones:
Can close this one now, yeah?

PR #243, opened 2021-04-03
new op CNT, counts down, stops at zero

attejensen:
#### What does this PR do?
Add new OP `CNT`, counts down, stops at zero

CNT x // should return x - 1, if the result is less than 0, return 0

Counting down and taking action when 0 is reached is something I use a lot, CNT makes it much more compact to do this.

#### I have,
* [x] updated `CHANGELOG.md`
* [x] run tests


scanner-darkly:
a couple of comments:

- it's aimed to provide a useful function but a narrow definition of that function - what if i want to count up instead of down? what if i want it to stop at 1 instead of 0?
- to use it as a counter, you still have to store the modified value: `A CNT A`. i think it would be more natural for it to behave as a variable where `CNT x` sets the start counter value and `CNT` modifies and returns that value. similar to what `O` variable would do if you set it up this way: `O.INC -1; O.WRAP 0; O.MIN 0`

attejensen:
Thanks, very good points indeed. Let me see if I understand you correctly: change CNT to works like O except it's local and has defaults of CNT.INC -1; CNT.STOP 0?

attejensen:
Hmmmm. I think I'll have to play with using `O` with `O.INC -1; O.WRAP 0; O.MIN 0` in actual musical examples and see how that goes. I have the feeling the "change on access" nature of `O` makes it awkward for this purpose...

attejensen:
I think I'll take this to the forum for a broader discussion, if that's OK with you?

scanner-darkly:
sounds good!

desolationjones:
Allow me to proselytize: I love `O` and I think it will work for you. In fact I love `O` so much that I'm tempted to make `O2` thru `O4` on my local branch. You can always set `O.INC 0` if you need to park it for multiple access, or assign it to a local variable.

attejensen:
Closing for now, will bring the discussion to lines...

PR #245, opened 2021-05-05
new OP hz (convert from v/oct to hz/v

attejensen:
#### What does this PR do?
Add new OP `HZ`, converts from v/oct to hz/v

https://llllllll.co/t/teletype-3-2-feature-requests-and-discussions/32052/369

(migth be one off, rounding "error")
HZ V 0 = V 1
HZ V 1 = V2
HZ V 2 = V4

This OP allows (along with inverted triggers) direct control of hz/v synths like the Korg MS.20

#### I have,
* [x] updated `CHANGELOG.md`
* [x] run tests


attejensen:
Hmmm, seems this is strangely incompatible with one of my private patches, TT boots with black screen, taking this down while I investigate....

scanner-darkly:
i've hit this issue too when trying to add a new feature, and wasn't able to resolve it or determine what causes it. it seems to be intermittent. debugging via ftdi didn't provide any helpful info as it would appear frozen at different points in the init routine. perhaps we're hitting some memory limits but i'm not sure how to tell if we are. 

cc @tehn

attejensen:
Memory limit was my initial thought as well, but I'm not sure what memory we're looking at. I suppose we have flash for the firmware and some ram besides that or? I tried decreasing SCENE_SLOTS, but that didn't change anything.

As you might know, one of my private patches is adding 8 scripts, for this to compile I have decreased SCENE_SLOTS to 24, so in this regard I'm close to the limit, still when flashing I see just above 50% memory, seems there are at least two different numbers in play.

Maybe we should hunt down all those int and int16_t in for loops and replace with uint8_t, I'm guilty here as well. Again a hunch, but isn't help-mode eating a good chunk with all these pages of text?

In any case, the problem turned out to be an old patch, that I frankly isn't even sure has ever worked in the first place, I accidentally enabled it, and thought the hz patch was to blame.

tehn:
the onboard help is getting pretty huge indeed. i'm not sure the best approach to audit the space allocation across different features, as much of it is a sort of user poll regarding what we should perhaps remove. people generally don't like feature removal :)

scanner-darkly:
i'm not 100% convinced it's memory related, just mentioned it as a possible cause. it could be that we still have enough memory for data structures, but maybe we're exceeding stack memory or something like that? i don't really know much about this area.. i tried analyzing the usage based on [sam's post](https://samdoshi.com/post/2016/10/avr32-ram-and-rom-usage/) but didn't really know what things to look for.

tehn:
@dndrks and i just checked over help mode, which neither of us had literally ever used.

1. we didn't know how to enter help mode. proposition: add a line at startup (after teletype version) that says "alt-h for help"
2. a command reference doesn't make sense to hold in memory. it's somewhat impossible to browse without a search at this point. i'd suggest we lose every page except the hotkey reference, which (to me) is integral given it's harder perhaps to remember how to navigate than to remember how to program teletype, especially if you just have the few commands you want in your brain. learning and reference really should happen with an auxiliary manual/device

if there's great desire for the help mode, we can look into dynamically pulling the help stuff from flash rather than active memory (or perhaps we do this? i didn't check the code)

i don't think we should have configurable help builds. my vote is we just remove it. (the TI/TO list of ops is just absurd to have onboard)

scanner-darkly:
@csboling implemented search functionality in help, it's ctrl-f / ctrl-r. i think help is useful as a reference, especially for something that has a long list of commands, like ti/to - this is where you more likely to need a refresher on the syntax.

in any case, i don't know if this would fix the issue we experienced - in my experiments i commented out all the grid code, and it didn't fix it. there is something else going on.

tehn:
ah, great to know there's a search! that improves my opinion.

and given i misunderstood that resource usage (of help particularly) was the problem, i apologize for the noise.

attejensen:
In any case,it would be interesting to know how many people are actually using the help. Seems we have different guesses on this :-) My guess is that it's mostly people who only touch the tt lightly and rarely. But obvious, if it does no harm, we might as well leave it there...

Now a search in the scene code is something I would really love.....

attejensen:
Let's continue this on lines:
https://llllllll.co/t/teletype-firmware-discussion/13961/96

PR #251, opened 2021-05-21
reduce type size of table data for less ROM. UNTESTED

trentgill:
#### What does this PR do?

Changes the types of data in the src/tables.c file. I made some guesses about which vals should be signed vs unsigned, but i reduced everything to the smallest storage type that would fit the data.

It saves ~1.5kB in ROM usage.

Every command that uses table data should be tested. I haven't been using TT enough to know what those commands are, so I'm not comfortable testing this.

Please don't merge until it has been tested.

#### I have,
* [ ] updated `CHANGELOG.md`
* [ ] run tests

### DIFF
--- a/src/table.c
+++ b/src/table.c
@@ -2,11 +2,11 @@
 
 // const int table_v[11] = {0, 3277, 6554, 9830, 13107, 16384, 19661, 22938,
 // 26214, 29491, 32768 };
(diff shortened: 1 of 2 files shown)


tehn:
good find re: assumptions about `int` sizes

attejensen:
The tables I created (table_hzv, table_n_b) looks fine!

PR #252, opened 2021-05-24
Customize NVRAM address to enable larger core flash program

### LINKED ISSUE #250: ROM is overflowing causing boot failures
#### Please describe the bug.

As described on [lines](https://llllllll.co/t/teletype-firmware-discussion/13961/102):

I did some investigation and I believe the issue is that the FLASH copy of the .data section (ie. statically initialized values of data) is overflowing into the .nvram section (where scripts etc are stored). This is evidenced by the fact that the .flash_nvram section has VMA address of 0x80040000 but LMA address of 0x80040be4 (on current main). This is from looking at module/teletype.lss after building the firmware.

This isn’t a problem in and of itself (the NVRAM is not using a full 128kB), but it becomes a problem because of how scenes etc are stored here 2. Note that the address of f resolves to 0x80040000.

Everything boots fine the first time and does the first-boot-initialization, overwriting the end of the .data section. So on the next boot, the firmware tries to copy a bunch of statically initialized values from ROM, but the values have been trashed by the re-initialization of scripts.

There needs to be some optimization of what is held in ROM. I’m not familiar with where all the chunky data is, but it’s over by 3kB in the current main branch, so something sizeable needs to be cut.

#### Please include the hash displayed at startup and the release of the software (e.g. "Teletype v2.0.1 5f838c9")

On current `main` branch, hash: 0x5acbc24.

1. Flash the firmware
2. Boots ok first time, writing the script data
3. Next boot fails, as a bunch of init values for static variables (saved in ROM) have been trashed.

#### Some initial ideas for where ROM can be saved

1. Table data is stored as `int`, but all data fits in `int16_t` or `uint16_t`: https://github.com/monome/teletype/blob/main/src/table.h
2. Help Mode is generally listed in pairs of `{name, usage}` with usage prepended by 2-4 spaces. It adds up! This indentation could be automatically inserted, rather than stored in a literal string.

trentgill:
#### What does this PR do?

Sets a custom symbol for the linker script, causing the NVRAM to be moved further into the flash address space. This was necessitated as the TT program has grown beyond the allotted 256KB flash reserved for it.

NVRAM is just a marker in the flash memory space which enables an additional interface for read/writing data. It does *not* require any special hardware support, and the location is arbitrary - the flash memory it controls is homogeneous.

TT uses the NVRAM address marker as the basis for all non-volatile data storage (Saved scenes, and calibration data). The 512KB of flash memory was previously split 256/256kB between program/data (the default). This PR changes the split to 312/200kB. The current usage of NVRAM is ~198kB, so if any substantial amount of data is added to the stored data, this variable may need to be updated.

https://llllllll.co/t/teletype-firmware-discussion/13961/102?u=galapagoose

Typical TT usage, and particularly, making sure saved Scenes & calibration data are working across power cycles.

fixes #250 

#### I have,
* [ ] updated `CHANGELOG.md`
* [ ] run tests

### DIFF
--- a/module/config.mk
+++ b/module/config.mk
@@ -250,7 +250,8 @@ CFLAGS = -fshort-enums -fno-common
 CPPFLAGS = -D BOARD=USER_BOARD -D UHD_ENABLE
 
 # Extra flags to use when linking
-LDFLAGS = -Wl,-e,_trampoline
+# NVRAM size may need to change if additional data is to be stored in scenes.
+LDFLAGS = -Wl,-e,_trampoline,--defsym=__flash_nvram_size__=200K
 
 # Pre- and post-build commands
 PREBUILD_CMD =


tehn:
gah. i just did this same thing, with a different route (made an additional `lds` file for teletype)

we can go this route instead, if we don't think any other bits of the `lds` will need modifying

scanner-darkly:
this might be a better option - since this setting might need to change in the future (if we free enough memory through refactoring and need to add something else to stored scenes), it’s easier to change it just in the main repo as opposed to having to change libavr32 as well

PR #257, opened 2021-05-25
new range OPs

attejensen:
#### What does this PR do?
Add new range OPs

https://llllllll.co/t/teletype-workflow-basics-and-questions/12392/1458

`INR lo x hi`  should be same as `&& < lo x < x hi`

#### I have,
* [x] updated `CHANGELOG.md`
* [x] run tests

NB: New Q OPs not in help_mode...


docs/ops/queue.toml
```diff
@@ -11,7 +11,7 @@ prototype = "Q.N"
 prototype_set = "Q.N x"
 short = "The queue length"
 description = """
-Gets/sets the length of the queue.
+Gets/sets the length of the queue. The length is 1-indexed.
```
scanner-darkly, line 14: not sure what this means? length is just the total number of elements in the queue, index doesn’t apply here, no?

docs/ops/queue.toml
```diff
@@ -22,3 +22,170 @@ description = """
 Getting the value the average of the values in the queue. Setting `x` sets the
 value of each entry in the queue to `x`.
 """
+
+["Q.CLR"]
+prototype = "Q.CLR"
+prototype_set = "Q.CLR x"
+short = "Clear queue"
+description = """
+Clear queue, set all values to 0, length to 1. If parameter `x` is provided, set first elements to `x`.
```
scanner-darkly, line 31: is it possible to have queue length of 0?

docs/ops/queue.toml
```diff
@@ -22,3 +22,170 @@ description = """
...
+
+["Q.RND"]
+prototype = "Q.RND"
+prototype_set = "Q.RND x"
+short = "Get random element/randomize elements"
+description = """
+Get a random element in queue.
+
+If `x` > 0, set all elements to a random value 0-`x`.
+If `x` < 0, swap two elements `-x` number of times.
```
scanner-darkly, line 74: would this be better as a separate op?

docs/ops/queue.toml
```diff
@@ -22,3 +22,170 @@ description = """
...
+short = "Sort all or part of queue"
+description = """
+Sort elements in queue.
+With no arguments, entire queue is sorted in accending order.
+
+If `x` > 0, sort elements from index `i` to the end of queue.
+If `x` < 0, sort elements from beginning of queue to index `-i`.
+IF `x` == 0, sort all elements.
+
+Index `i` is 0-indexed.
```
scanner-darkly, line 90: is it 1-indexed in other places? if so, we should keep it consistent

src/ops/queue.c
```diff
@@ -70,3 +158,384 @@ static void op_Q_N_set(const void *NOTUSED(data), scene_state_t *ss,
...
+        for (int16_t i = rnd; i < 0; i++) {
+            a = rand() % q_n;
+            b = rand() % q_n;
+            tmp = q[a];
+            q[a] = q[b];
+            q[b] = tmp;
+        }
+    }
+    else {
+        // what to to with rnd = 0????
```
scanner-darkly, line 278: old comment? the doc says it will do nothing

docs/ops/queue.toml
attejensen: Yeah, I guess you're right. Was just trying to make a note to set it apart from the 0-indexed nature of the rest of them...

docs/ops/queue.toml
attejensen: No, mainly to keep in line with the general though of old Q.

docs/ops/queue.toml
attejensen: Could be...

docs/ops/queue.toml
scanner-darkly: i can’t really comment on this as i don’t use the queue, but to me it doesn’t make sense to have a queue that always has at least 1 element. but if that’s what the original implementation does then makes sense to keep it that way.

scanner-darkly:
looks like `Q.CLR` failed the tests

docs/ops/queue.toml
desolationjones: If you clear the queue then add an element, does the residual 0 get pushed into the next spot, or overwritten?

src/ops/queue.c
attejensen: Basically RND is doing a few things:

Q.RND // return random element
Q.RND x (x>0) // set elements to random value between 0 and x
Q.RND x (x<0) // swaps elements -x times 

The comment was to remind the reader (you?) that Q.RND 0 doesn't do anything.

I realize it's a but multi purpose, OTOH it's filling gaps that would otherwise be empty, like it doesn't make sense to rote a negative amount of times, and filling with random elements < 0, could be accomplished with Q.MUL -1.

If you feel it's messy, any chance you could suggest a split of this OP?

desolationjones:
> looks like `Q.CLR` failed the tests

Speculation: it accepts command state pointer argument but doesn't make use of it. Is that significant?

src/ops/queue.c
scanner-darkly: > The comment was to remind the reader (you?) that Q.RND 0 doesn't do anything.
yeah and i think it's helpful, it was just the question marks that were confusing to me, as they imply this is an unresolved question.

scanner-darkly:
shouldn't matter, it's usually a mismatch in the number of parameters and whether or not it returns a value

desolationjones:
> shouldn't matter, it's usually a mismatch in the number of parameters and whether or not it returns a value

Sure enough, looks like the "returns a value?" flag is set to "true".

docs/ops/queue.toml
attejensen: It depends on GRW (grow)

Q.CLR // now Q is 1 long and all 0s
Q.CLR 42 // now Q is 1 long, first element (idx 0) is 42, the rest are 0
Q 1 // length is still 1, first elemtn is 1, 2nd element is 42, but "past the length border"
Q.GRW 1 // now queue grows automatically
Q 2 // now Q is 2 long, first 2, 2nd 1. A third element 42 is "behing the length border", the rest are 0

Yeah, I didn't dream up the org Q, took me ages to figure out (and document) how that worked, quirky as is it, I'm just trying to extend what's already there.

Not saying RTM at all, but I propose a read through the examples in the docs, I hope it's clearer after that :-)

scanner-darkly: i'd say this is the case where community input is more valuable than any feedback i would be able to provide :)

attejensen: I agree, the original post didn't stir anything up (except a few enthusiastic comments)
https://llllllll.co/t/teletype-3-2-feature-requests-and-discussions/32052/345

Could we put it in the beta as-is and get feedback once people have tried it out themselves?

attejensen:
Any hints on understanding what Travis is complaining about? Maybe I'm just stupid, but I see no clear indication on what he's mad about.....

docs/ops/queue.toml
desolationjones: Ok I did a re-read of the docs and it's much clearer to me now. The only thing which is not mentioned anywhere is whether the queue-modifying OPs only affect in-bound values?

scanner-darkly: sounds like a question for @tehn 
(my personal view is - we don't want to end up with a scenario similar to patterns where start/end/length usage is confusing and inconsistent, and adding new ops will make it more difficult to change the approach if needed. but i shouldn't really comment as i don't use the queue and therefore my opinion here might be just creating noise)

desolationjones:
You have Q.SUM it as GET/SET but only defined a GET

attejensen:
Sure everything build locally, I must be messing something up out of the ordinary...

desolationjones:
I think Travis performs a similar thing to https://github.com/monome/teletype/#tests i.e.
`cd tests`
`make`

attejensen:
Thanks @desolationjones , didn't know about that. Local tests pass now...

attejensen:
Grr, had to hack travis to get to this, maybe I'm missing a verbose flag, but travis didn't tell me where it had floating poing exception....

docs/ops/queue.toml
attejensen: I will go over the docs again, but everything respects in-bound values. EXCEPT direct indexing Q.I and copy from pattern

PR #258, opened 2021-05-29
quantize OPs use 1V/OCT and bitwise N.B scales

desolationjones:
#### What does this PR do?
Modifies beta OPs `QT.S`, `QT.CS`, and `QT.B` to accept & output 1V/oct signals instead of note numbers. Changes `QT.B` to use scale defined and stored by `N.B`. Creates `QT.BX` which references a stored scale from `N.BX`. Refactors `QT.S` and `QT.CS` to use common bitwise scale quantization functions.

Use 1V/oct signals into each quantizer and use output with a VCO. Compare to keyboard or something. Pitch accuracy isn't super amazing though since it's all using the 1st octave of `table_n` plus octave multipliers.

#### I have,
* [x] updated `CHANGELOG.md`
* [x] run tests


scanner-darkly (approved):
looks good, can't comment on the ops themselves or the math behind since i don't really understand what they do.

one comment - there is a commented out function in `maths.c`, should that be removed?

scanner-darkly (approved):

PR #260, opened 2021-06-02
Load scene while keeping current pattern state

eimhin:
#### What does this PR do?

This op loads a scene without loading the scenes pattern data, similar to `SCENE.G` for grid configuration.

Create two scenes with different pattern data. Load the first scene. Use `SCENE.P` to load the second scene. Confirm that the pattern data has stayed the same, instead of loading pattern data from the second scene.

The idea behind this feature is to be able to jam with the same set of pattern data across multiple scenes.

#### I have,
* [x] updated `CHANGELOG.md`
* [x] run tests


scanner-darkly (approved):

scanner-darkly:
nice work!

PR #264, opened 2021-06-10
live mode dashboard / clean up

scanner-darkly:
#### What does this PR do?

- live mode dashboard - the ability to display custom text and print values on live mode screen
- new `alt-d` shortcut for the above
- new ops for the above `LIVE.DASH`, `LIVE.VARS`, `LIVE.GRID`, `LIVE.OFF` and `PRINT`
- general clean up: clang format on everything, missing commas in help mode, missing stuff in whats_new

see the doc for the `LIVE.DASH` op

#### I have,
* [x] updated `CHANGELOG.md`
* [x] run tests


scanner-darkly:
the tests were definitely passing locally, not sure why travis is complaining, will take a look

scanner-darkly:
the tests are failing on ops added in #PR257. @attejensen can you take a look?

scanner-darkly:
so looks like clang format broke some of the ops - which is why the tests passed for me locally - i ran clang format after. should be fixed now.

desolationjones:
> 
> 
> so looks like clang format broke some of the ops - which is why the tests passed for me locally - i ran clang format after. should be fixed now.

Won't these break every time they are formatted? Perhaps we should wrap that section in a `// clang-format off`.

tehn (approved):
this is super super rad

scanner-darkly:
@desolationjones `math.c` already has `clang-format off` but it still modified it. not sure why..

PR #271, opened 2021-07-08
don't allow out-of-range scenes to be selected from SCENE OP

attejensen:
#### What does this PR do?
Protect scene from being out of range (negative or larger than SCENE_SLOTS -1)

SCENE -1 (or smaller) should change to scene 0
SCENE 32 (or larger) should change to scene 31

#### I have,
* [ x] run tests

### DIFF
--- a/src/ops/controlflow.c
+++ b/src/ops/controlflow.c
@@ -1,4 +1,5 @@
 #include "ops/controlflow.h"
+#include "../module/flash.h"
 
 #include "helpers.h"
 #include "random.h"
@@ -221,6 +222,9 @@ static void op_SCENE_get(const void *NOTUSED(data), scene_state_t *ss,
 static void op_SCENE_set(const void *NOTUSED(data), scene_state_t *ss,
                          exec_state_t *NOTUSED(es), command_state_t *cs) {
     int16_t scene = cs_pop(cs);
+    if (scene < 0) { scene = 0; }
+    if (scene >= SCENE_SLOTS) { scene = SCENE_SLOTS - 1; }
+
     if (!ss->initializing) {
         ss->variables.scene = scene;
         tele_scene(scene, 1, 1);
--- a/src/ops/controlflow.c
+++ b/src/ops/controlflow.c
@@ -234,6 +234,9 @@ static void op_SCENE_set(const void *NOTUSED(data), scene_state_t *ss,
 static void op_SCENE_G_get(const void *NOTUSED(data), scene_state_t *ss,
                            exec_state_t *NOTUSED(es), command_state_t *cs) {
     int16_t scene = cs_pop(cs);
+    if (scene < 0) { scene = 0; }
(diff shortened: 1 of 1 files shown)


scanner-darkly:
this was already addressed in https://github.com/monome/teletype/pull/270

attejensen:
I don't agree that exposing is better than protecting. I can't think of anywhere else where we prefer to show the user their mistake rather than correct it for them. To me this forgiving style is one of the fundamental features of tt.

IOW: At least the behavior feels illogical to me, esp since we're clamping almost as per default, making this one feel like a bug...

scanner-darkly:
i think you misunderstood what i meant - i don't suggest not protecting `SCENE` ops and exposing a user error in the form of teletype crashing. what i meant is - we could clamp values to the expected range, but to me not doing anything when you execute `SCENE 100` is better than interpreting it as `SCENE 31`, as otherwise it appears my script is working - except that it's doing something that i didn't intend it to do. if you do want to clamp your `SCENE` parameter you can do that explicitly with `LIM` op.

this is not a newly introduced behaviour - it follows exact same approach as `SCRIPT`:
https://github.com/monome/teletype/blob/main/src/ops/controlflow.c#L258

for the record, it's one of the things not done consistently in teletype - pattern indexes are clamped, for instance, but for the aforementioned reasons i think we should follow what `SCRIPT` does.

regardless of what we do, the underlying functions should have their own protection in place, which is what the following fixes do:

https://github.com/monome/teletype/pull/270/files#diff-3a4cca72d1d39e12a389df0b15529f7f96d0fc91ceae2fa998063f5b8d564f9eR1068

https://github.com/monome/teletype/pull/270/files#diff-1b03fbdca6cb0fa3699bc4ec8a3c31b1b1c30b0d4b74b33825e9eea0c5d7c5baR122

attejensen:
I didn't misunderstand. I still think it's better to clamp everywhere, really don't see why SCENE and SCRIPT are any different from say P or Q. And what bothers me the most is the bonkers state where SCENE can return a negative number.

But I'll rest my case here, and just patch locally...

scanner-darkly:
`SCENE` returning a negative value sounds like a separate bug, what are the repro steps for that?

i presented some specific pros of not clamping - happy to discuss specific cons. and we should take this discussion to the thread on lines so others can chime in too.

attejensen:
SCENE 0 \\ now on scene 0
SCENE \\ correctly returns 0
SCENE -1 \\ woot
SCENE \\ returns -1
SCENE 0 \\ back to well defined state
SCENE.G -1 \\ oops again
SCENE \\ -1 again
SCENE 0 \\ well defined state
SCENE.P -1 \\ -1 again
SCENE \\ and it returns -1 yet again

attejensen:
I stated a new topic on lines here:
https://llllllll.co/t/teletype-clamping-vs-no-change/46957

PR #277, opened 2021-09-22
Making PROB slightly more useful

djwglpuppy:
#### What does this PR do?

Adds Functionality to the following

- `PRB` is now an alias of `PROB`

Having the extra character helps when createing complex patches based on Probability

`PROB` can now use the `ELSE` statement

A common use case is to use a Bernoulli gate.  What this means is that it either goes to A OR to B based on a weighted coin toss

example/. (60% chance of triggering 1, otherwise trigger 2)

```
PRB 60: TR.P 1
ELSE: TR.P 2
```

See example above

#### I have,
* [ X] updated `CHANGELOG.md` and `whats_new.md`
* [X ] updated the documentation
* [X ] run `make format` on each commit
* [X ] run tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,7 @@ __pycache__/
 /docs/cheatsheet/*
 !/docs/cheatsheet/cheatsheet.tex
 /docs/testServe/
+
+# Jet Brains
+.idea
+*.iml
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,9 @@
 - **FIX**: delay when opening docs
 - **FIX**: `PROB 100` would only execute 99.01% of the time.
 - **FIX**: some `G.FDR` configurations caused incorrect rendering in grid visualizer
-
+- **NEW**: `PROB` accepts an `ELSE` condition
+- **NEW**: alias `PRB` for `PROB`
+- 
 ## v4.0.0
 
 - **FIX**: `LAST SCRIPT` in live mode gives time since init script was run
--- a/docs/ops/controlflow.toml
+++ b/docs/ops/controlflow.toml
@@ -4,7 +4,7 @@ short = "if `x` is not zero execute command"
 description = """
 If `x` is not zero execute command
 
-#### Advanced `IF` / `ELIF` / `ELSE` usage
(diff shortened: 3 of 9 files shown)


cristianvogel:
Branching off prob is a good idea. Nevertheless the example you give is easily done with `TR.P + 1 TOSS `

tehn:
this is a nice addition, thank you.

i'm a little concern that the `IF / PROB` pattern in the documentation makes things a little confusing for first time readers, and it'd be better to mention PROB as an addendum to the IF/ELSE section.

djwglpuppy:
thanks for reviewing thus far.   I agree with the docs, it felt correct putting the `PROB` along side the the other conditional, but it comes across a little clunky.   I will adjust.

also @tehn I am about to do another PR based on a new sibling operator to `PROB` called `WTOSS` (`WTOSS x` where `x` is the weight from 0 to 100 (alias `WT`)) which is a weighted toss.   It runs in a similar fashion but instead returns a `1` or `0` instead of being a conditional.  I have been doing it manually all the time and thought it would be easier to just do stuff like

```
DEL ? WTOSS 90 300 3000: TR.P 1
```

or cascading weighted tosses based on this PR

```
PRB 50: TR.P 1
ELIF WT 70: TR.P 2
ELSE: TR.P 3 
```

Since they are both covering weighted probabilities, should I just add this new `OP` to this PR or do a new one?

djwglpuppy:
@tehn I added a new OP previously mentioned to this.   I typically try to keep my PRs to a singular case, but I decided to add on since they are so similar in nature for working with weighted probabilities.   I just built and tested these and really hope it makes it in the next release because I already see myself using this new stuff a ton.   Thankfully it was fairly easy to implement with my abysmal c knowledge.  Please lemme know if anything can be done more efficiently.

Refer to the comment above for testing `WTOSS` / `WT`

djwglpuppy:
I put the PR out in the forum and some people made some interesting points about `PROB` sitting between an `IF` / `ELSE` that works for their workflow.    Would be be better to split this into a new [sub]OP like `PROB.B`? (`PRB.B`) (B is for Bernoulli) That way existing `PROB` does not hurt their workflow.

https://llllllll.co/t/a-draft-of-some-prob-tweaks-i-made-for-teletype/48956/5

PR #278, opened 2021-09-25
Buffer each command line in process_command

beels:
#### What does this PR do?

This change stores the currently executing command in a local buffer during each call to `process_command`.  This ensures that when the scene changes, the operation map in `process_command` still points into the same code upon which it was based.  In this way we avoid undefined behavior when the global scene data is replaced while a command is running (i.e., when `SCENE` is executed).  Changing scenes from while code is running is still full of surprises, but they are well-defined.

https://llllllll.co/t/teletype-scene-x-op-users-of-that-op-chime-in/45833/23?u=3-foot-1

Run the scenes in the cited lines thread before and after applying the PR.  Before the change, A will be 9 and C will be 7.  After the change, A will be 18 and C will remain unchanged.

Extensive discussion in the_dev_zone on the Teletype Studies discord.

#### I have,
* [X] run tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v4.0.x
 
+- *FIX*: cache currently-running commands to avoid corruption during SCENE ops.
 - **FIX**: delay when opening docs
 - **FIX**: `PROB 100` would only execute 99.01% of the time.
 - **FIX**: some `G.FDR` configurations caused incorrect rendering in grid visualizer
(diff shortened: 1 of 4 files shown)


beels:
No change to documentation or help mode, since this fix does not change documented behavior.

I don't have clang format, so can't run `make format` at this time.

beels:
Got clang format working.

scanner-darkly (approved):

scanner-darkly:
yeah i think this makes sense!

PR #279, opened 2021-10-10
added Q.POP, Q.FND and Q.REM

attejensen:
#### What does this PR do?
Adds three new OPs
Q.POP: pop first element or element at index from list (stack)
Q.FND: return index of first element with value x, -1 if not found
Q.REM: remove first element with value x, return index of removed element, -1 of none removed

#### How should this be manually tested?
Q.GRW 1 // enable grow
Q.CLR 2 // put 2 on stack
Q 4         // put 4 on stack
Q 6         // put 6 on stack
Q 8         // put 8 on stack, now stack is | 8 6 4 2 |
Q.POP   // pop first element, will return 8, stack is now | 6 4 2 |
Q.POP 1 // pop element at index 1, will return 4, stack is now | 6 2 |
Q.FND 1  // find element with value 1, none is found returns -1
Q.FND 2  / find element with value 2, found at index 1, return 1, stack is unchanged
Q.REM 2 // remove first element with value 2, return index of removed element, will return 1, stack is now | 6 |
Q.REM 1 // remove first element with value 1, not found, will return -1, stack is unchanged

#### Any background context you want to provide?
This makes it possible to use Q as stack. Searching and removing elements by value are handy tools.

#### I have,
* [x] updated `CHANGELOG.md` and `whats_new.md`
* [x] updated the documentation
* [x] updated `help_mode.c` (if applicable)
* [x] run `make format` on each commit
* [x] run tests

PR #280, opened 2021-10-12
Create a simple Actions workflow for tests

Dewb:
Saw the comment about GitHub Actions in #279, I've been doing a bunch of Actions stuff lately (including building TT in VCV Rack) so I thought it would be pretty quick to whip one up.

I meant to open this PR against my own fork first, was opened here prematurely -- here's the Action in action:
https://github.com/Dewb/teletype/runs/3876503376?check_suite_focus=true

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -57,3 +57,8 @@ jobs:
         run: |
           cd module
           make
+      - name: Upload teletype.hex to artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          path: module/teletype.hex
+          name: teletype.hex


Dewb:
Split the test and simulator jobs, and added a third job to set up the toolchain and build the firmware.

> i assume it doesn't need to run op_enums.py?

It could! But no, that's not currently part of the build process (or the Travis checks, which is what I was modeling this on.)

Dewb:
Ah yes, `op_enums.py` is mentioned in the README, but not the PR template or any of the build scripts. 

Since python is already a docs dependency, might be worth just removing that file from source control and building it as part of the Makefile, like the ragel-generated source files?

scanner-darkly:
> but it's not mentioned in the PR template or any of the build scripts
likely because you have to run it in order to add a new op, so there is no need to check if you actually did.

re: makefile - for whatever reason it doesn’t work on my windows box so i execute build steps manually (ragel too)

PR #285, opened 2022-05-16
Insert cv and param in tracker mode

mikeyford:
#### What does this PR do?
Adds data to patterns using CV input and param values, when in tracker view. Respects settings for IN.SCALE and PARAM.SCALE.

Go into tracker mode, move param knob while pressing `p`. Supply CV at input while pressing `i`

I was looking for quicker way to create sequences from a keyboard with CV out.

#### I have,
* [X] updated `help_mode.c` (if applicable)
* [X] run tests

### DIFF
--- a/module/help_mode.c
+++ b/module/help_mode.c
@@ -91,7 +91,9 @@ const char* help1[HELP1_LENGTH] = { "1/16 HELP",
                                     "SH-S|SET START",
                                     "SH-E|SET END",
                                     "ALT-L,S,E|JUMP",
-                                    "SHIFT-2|SHOW/HIDE TURTLE" };
+                                    "SHIFT-2|SHOW/HIDE TURTLE",
+                                    "I|SET TO CV INPUT",
+                                    "P|SET TO PARAM KNOB" };
 
 #define HELP2_LENGTH 25
 const char* help2[HELP2_LENGTH] = { "2/16 VARIABLES",
--- a/module/help_mode.c
+++ b/module/help_mode.c
@@ -20,7 +20,7 @@
 
 // clang-format off
 
-#define HELP1_LENGTH 71
+#define HELP1_LENGTH 73
 const char* help1[HELP1_LENGTH] = { "1/16 HELP",
                                     "[ ] NAVIGATE HELP PAGES",
                                     "UP/DOWN TO SCROLL",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - **FIX**: `PROB 100` would only execute 99.01% of the time.
 - **FIX**: some `G.FDR` configurations caused incorrect rendering in grid visualizer
 - **FIX**: fix `EX.LP` not returning correct values
+- **NEW**: `i` and `p` insert values from CV input and param knob in tracker mode
 
 ## v4.0.0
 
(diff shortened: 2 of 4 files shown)


module/help_mode.c
```diff
@@ -91,7 +91,9 @@ const char* help1[HELP1_LENGTH] = { "1/16 HELP",
                                     "SH-S|SET START",
                                     "SH-E|SET END",
                                     "ALT-L,S,E|JUMP",
-                                    "SHIFT-2|SHOW/HIDE TURTLE" };
+                                    "SHIFT-2|SHOW/HIDE TURTLE",
+                                    "I|SET TO CV INPUT",
+                                    "P|SET TO PARAM KNOB" };
```
tehn, line 96: also need to modify HELP1_LENGTH const

tehn (changes requested):
super cool feature, thanks for adding!

tehn:
not sure how i missed the existing functions, thanks for the heads up

scanner-darkly:
to clarify - currently there is no way to record the input value, just the param value, and i can see the benefit of using PARAM.SCALE in addition to (or maybe as a replacement for?) the predetermined 0..31 / 0..1023 ranges.

mikeyford:
My primary motivation here was entry from CV. I added the param input option to use the `PARAM.SCALE` values mostly since I was here too. It feels more flexible/useful to me than the existing ranges, but I also don't feel at all strongly about it, so happy to drop it if that's preferred.

scanner-darkly:
i think both are useful additions since they offer different functionality!

this does bring up a question of using letter shortcuts in tracker view - so far we don’t have any, and i don’t see why there shouldn’t be some, so the question becomes whether we should add a bunch of others to improve tracker functionality (and maybe make it more of an actual tracker), in which case it would be good to consider what else could be added and allocate letters accordingly.

scanner-darkly:
are you on teletype study discord? we have a dev channel there which is a great place to get quick feedback!

tehn (approved):

tehn:
agreed with @scanner-darkly --- before introducing letter input to the tracker screen it'd be good to assess letter assignments more broadly

scanner-darkly:
perhaps something like shift-p and shift-i for now?

desolationjones:
I could get behind just changing the existing shortcuts ctrl-alt and ctrl-shift to paste param and in, respectively, scaled to their individual scaling OPs.

Currently:

<html><body>


ctrl-alt | insert knob value scaled to 0..31
-- | --
ctrl-shift | insert knob value scaled to 0..1023


</body>
</html>

Proposed:
<html><body>


ctrl-alt | insert PARAM knob value scaled to `PARAM.SCALE`
-- | --
ctrl-shift | insert `IN` CV value scaled to `IN.SCALE`


</body>
</html>

@mikeyford feel free to pop by the [discord study group](https://discord.gg/WsUAhqSb) to discuss!

scanner-darkly:
i think it's best to keep the existing ones - there might be folks that are used to the old shortcuts, and it would create confusion if people reference the manual but are still on an older version.

mikeyford:
Apologies. Haven't had time to follow up on this yet, and I'm away from my Teletype for the next few weeks. 

I am planning to join the Discord and discuss this change there once I have time. Also interested/motivated to add further functionality to tracker mode, after collecting some input on the Discord.

Issue #104, opened 2017-09-24
scenes with `#` in description don't load properly


scanner-darkly:
if a scene has a `#` character in the description when such scene is loaded everything after the `#` will not be read, including all of the scripts.

burnsauce:
Given that # is reserved in the teletype scene file syntax, I presume that the fix should be to disallow # in the scene description?

samdoshi:
Sounds like that's what we need to do until we rewrite the USB code.

It should be as simple as adding a branch to match `#` in [`process_preset_w_keys`](https://github.com/monome/teletype/blob/master/module/preset_w_mode.c#L37).

scanner-darkly:
perhaps a better option would be to allow entering it but replace it with something else at that point (a space?), not just ignore it - otherwise it might look like keyboard is not working, and there is no messaging provided to the user to clarify what's going on.

cmcavoy:
I looked for existing C serializer libraries we could leverage. I couldn't find any that seemed usable, but did see mention of using `/* */` for comments in this [implementation](http://www.happyponyland.net/cserialization/readme.html). Could we use an alternate comment format in TT scripts?

It's going to take me a while to understand the constraints around embedded development, but my first idea reading this thread was to leverage an existing serializer, doesn't sound like it's possible here - the embedded world is fun to learn.

samdoshi:
I think discussions here are being overtaken by discussions at:

https://llllllll.co/t/teletype-firmware-usb-disk-mode-interface/9442

We are currently contemplating switching from a text file format to a binary format.

jlmitch5:
> perhaps a better option would be to allow entering it but replace it with something else at that point (a space?), not just ignore it - otherwise it might look like keyboard is not working, and there is no messaging provided to the user to clarify what's going on.

My guess would be someone would be typing the `#` key because they are trying to document the script. i.e.:

```
#1:
foo

#2:
bar

#3:
baz
```

In that situation I don't really think replacing `#` with anything would really make sense to the user.  My gut is that just not responding to the character press is probably the least invasive action to take pre-changing all the big, serializer changes.

cmcavoy:
Can someone point to the serialization / deserialization code? Just curious to see it but have had trouble finding it in the source.

scanner-darkly:
https://github.com/monome/teletype/blob/master/module/usb_disk_mode.c

PR #287, opened 2022-05-17
Proposal: separate scene serialization logic from USB disk logic, add scene serialization tests

Dewb:
### What does this PR do?

This PR separates the scene parsing code and the USB disk handling code, so scenes could theoretically be loaded from and saved to other sources besides the USB drive. 

It creates an abstract interface for scene serialization and leverages it to add unit tests for scene I/O.  When invoked in the USB disk mode, the sequence of operations is more or less unchanged; this doesn't substantially alter the user-visible behavior.

The included presets are also updated to the latest syntax (newer II ops, PRE colon spacing) so tests can round-trip them and compare the output against the original.

This also includes fixes for two bugs discovered while doing this work:
   * inits scene calibration data to reasonable values to avoid divide-by-zero in `process_tests` on some x86 hardware.
   * enlarges the text string buffer used for `WRITE.../READ...` in USB disk mode, avoiding overflowing the buffer and potentially smashing neighboring stack vars.

### How should this be manually tested?

* Load a USB stick with a selection of scenes in `tt*.txt`
* Insert USB stick into teletype, confirm WRITE/READ cycle completes
* Remove, attach keyboard, confirm scenes are loaded correctly
* Re-insert USB stick, repeat WRITE/READ cycle
* Insert USB stick into computer, compare saved files to originals with something like the below command; files should be the same 
   `for f in tt*.txt; do git diff $f ${f%.*}s.txt; done`

### Any background context you want to provide?

The original motivation for this was to be able to use the scene parser in VCV Rack, but also with an eye to benefit hardware development down the road. Hopefully this makes it a little easier to contemplate improvements like:
* CRLF line ending support
* #104 
* scene loading over i2c?
* new USB disk management UI??

### I have,
* ~~updated the documentation~~
* ~~updated `help_mode.c` (if applicable)~~
* [X] run tests


module/usb_disk_mode.c
```diff
@@ -25,17 +24,31 @@
...
+}
+
+bool tele_usb_eof(void* self_data) {
+    return file_eof() != 0;
+}
 
+// usb disk mode entry point
 void tele_usb_disk() {
-    char input_buffer[32];
+    char text_buffer[40];
```
Dewb, line 51: this is the fix to the `strcat` overflow. I noticed it because I initially put some additional local variables before this line, and they got corrupted by the end of the write cycle. I'm not sure what the consequences of this overflow would be, depends on the AVR32 stack frame layout, but it could theoretically mangle the return pointer, or stackvars in the calling function (like the saved irq flags?)

module/usb_disk_mode.c
```diff
@@ -107,94 +122,13 @@ void tele_usb_disk() {
                 continue;
             }
```
Dewb, line 124: the following code is moved mostly verbatim to scene_serialization.c, just with calls to `file_write_buf`, etc. replaced with calls to the serializer interfaces.

presets/tt06.txt
```diff
@@ -15,29 +15,24 @@ MUST CONNECT VIA CABLE
...
-II MP.FREEZE 1
 
 #6
-II MP.UNFREEZE 1
 
 #7
-L 1 8 : II MP.MUTE I
 
 #8
-L 1 8 : II MP.UNMUTE I
```
Dewb, line 40: the MP ops above are no longer supported.

tests/serialize_scene_tests.c
```diff
@@ -0,0 +1,191 @@
...
+        }
+        if (ca == '\n') { line++; }
+    }
+
+    if (feof(a) != feof(b)) { return -1; }
+
+    return 0;
+}
+
+TEST test_round_trip_file(char* filename, char* tempfile) {
```
Dewb, line 118: Looking at the tests might be the fastest way to get an overview of what's going on here.

tehn:
this looks good, thanks for taking it on! i've done a surface review but i'll dig deeper in the coming days.

tehn (approved):

scanner-darkly:
will take a look in the next couple of days as well!

scanner-darkly (approved):

scanner-darkly:
thank you for doing this, amazing!

Issue #291, opened 2022-10-12
M.RESET is required when updating metro rate


scanner-darkly:
#### Please describe the bug.

when metro rate is changed and the remaining time is longer than the new interval, it will set the remaining time to the interval, so the next time it triggers exactly after the specified interval, as expected. but if the remaining time is shorter it will allow it to complete before using the new interval, resulting in having to also execute `M.RESET` in order to use the new interval immediately.

#### Please include the hash displayed at startup and the release of the software (e.g. "Teletype v2.0.1 5f838c9")

this is still a bug as of 4.0 release

#### What steps are required to reproduce the bug?
1. set metro rate to 1 sec and output a trigger on each metro tick
2. change metro rate to 2 sec when it's in the 2nd half of the metro interval

actual result: it will finish the remaining interval before using the new 2 sec interval
expected result: it should apply the new interval immediately

desolationjones:
Can we close this issue now?

PR #292, opened 2022-10-12
reset M timer when changing metro rate

scanner-darkly:
#### What does this PR do?

fixes issue #291 

n/a

see #291 

#### I have,
* [x] run tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - **IMP**: scene load/save code refactor, add scene load/save tests
 - **IMP**: fader ops now support up to four faderbanks
 - **NEW**: new Disting EX ops: dual algorithms, `EX.M.N#`, `EX.M.NO#`, `EX.M.CC#`
+- **FIX**: reset M timer when changing metro rate
 
 ## v4.0.0
 
--- a/docs/whats_new.md
+++ b/docs/whats_new.md
@@ -10,6 +10,7 @@
 - **IMP**: scene load/save code refactor, add scene load/save tests
 - **IMP**: fader ops now support up to four faderbanks
 - **NEW**: new Disting EX ops: dual algorithms, `EX.M.N#`, `EX.M.NO#`, `EX.M.CC#`
+- **FIX**: reset M timer when changing metro rate
 
 ## v4.0.0
 
--- a/module/main.c
+++ b/module/main.c
@@ -992,7 +992,7 @@ void tele_metro_updated() {
         metro_timer_enabled = false;
     }
     else if (metro_timer_enabled) {  // just update the time
-        timer_reset_set(&metroTimer, metro_time);
+        metroTimer.ticks = metro_time;
     }
 
     if (metro_timer_enabled && ss_get_script_len(&scene_state, METRO_SCRIPT))
--- a/module/main.c
+++ b/module/main.c
@@ -992,7 +992,7 @@ void tele_metro_updated() {
         metro_timer_enabled = false;
     }
     else if (metro_timer_enabled) {  // just update the time
-        timer_set(&metroTimer, metro_time);
+        timer_reset_set(&metroTimer, metro_time);
     }
 
     if (metro_timer_enabled && ss_get_script_len(&scene_state, METRO_SCRIPT))


desolationjones:
Doesn't this break the ability to change metro rate at an arbitrary time without affecting its phase?

If I understand the bug report correctly, there's a few separate confusing and possibly broken metro behaviors. I'm using this line in M to visualize:
`PN.NEXT 0 M; PN.NEXT 1 LAST $`
Now every time the metro fires, you can see how long its period was set at that moment, and how long it actually waited since last fire. Changing between `M 1000` and `M 3000` in Live screen at arbitrary times, I get results as below (in VCV Rack, hopefully real hardware is similar):
```
1000 1000
1000 1000
3000 1000
3000 3000
1000 2464
1000 1000
1000 1000
```
Currently there are instances where lengthening the period doesn't affect the current cycle (phase is preserved but expected action is delayed), and instances where shortening the cycle causes the timer to be reset at that moment (phase can be ruined).

I think your change would be equivalent to using `M 1000; M.RESET` and `M 3000; M.RESET` in my example above. Results:
```
1000 1000
1000 1000
3000 3184
3000 3000
1000 3209
1000 1000
```
This still has the issue where you can ruin the phase of your metro by making changes to `M` at arbitrary times. Maybe this can be avoided by using a dummy variable which is only assigned to `M` when metro fires:
```
PN.NEXT 0 M; PN.NEXT 1 LAST $
M X; M.RESET
```
Results of setting `X 1000` and `X 3000` at arbitrary times:
```
1000 1000
1000 1000
3000 3000
3000 3000
1000 1000
```
But now we're stuck waiting a whole tick for metro period to be adjusted.

I think there's no "one right answer" so maybe there could be a modal behavior OP:
Mode 0: Classic Behavior
Mode 1: Setting M resets timer
Mode 2: Setting M has no effect until metro fires or `M.RESET` is used

Thoughts?

scanner-darkly:
so the current issue is that the behaviour is inconsistent when changing the metro rate. let's say A is the old interval and B is the new interval. the current behaviour:

- if the remaining time is less than B, it will allow it to finish. so it completes A, then the next interval will be B
- if the remaining time is more than B, it will set it to B. so we get an interval that is some arbitrary value in the B..A range, then the next interval will be B

indeed, my change is the same as executing `M.RESET` after setting the new interval which results in the more predictable behaviour but you're right in that it might not always be desirable.. what if instead of using different modes we did the following:

- we change `M` to only set the new interval, the current interval will be allowed to always complete
- we add `M.SET` which would be the equivalent of `M x; M.RESET` and would have the meaning of "apply the new interval immediately"

desolationjones:
Sounds good! The only inaccessible behavior would be the current behavior. I'm having a hard time thinking of what kind of scripts (if any) that may break.

scanner-darkly:
yeah i doubt there are scripts that depend on the current behaviour unless you have some really elaborate logic on when M changes happen.. and the new behaviour would be consistent with the current behaviour for at least some cases.

i think i'll leave `M.NOW` out as you could simply do `M x; $ 9` (which will have the same effect as `M x; M.RESET; $ 9`

scanner-darkly:
i added `M.SET` and played with it and i think it's actually even more confusing because if you do it as some arbitrary point between 2 metro ticks, the current interval will be B + some portion of A, which doesn't seem useful/intuitive. another option would be to change the remaining time so that the current interval equals B, but if the elapsed time is greater than B, we're still going to get an interval that is neither A nor B.. 

simply updating the metro interval starting with the next tick seems more intuitive, and if you want the new interval to be applied immediately you can still do `M x; M.RESET`. what do you think?

desolationjones:
Agreed, the existence of `$ 9` and `M.RESET` covers those use cases.

Another possible metro scheme I hadn't considered is to treat the metro like an LFO, and preserve phase at the moment when the period is adjusted. E.g. if you're 500 ms into an `M 1000` cycle, and you assign `M 3000`, the metro timer would be updated from 500 to 1500 ms.

It's perhaps an aesthetically attractive solution and it works consistently regardless of whether `M` is getting bigger or smaller. However, I think my preference remains "update `M` only at metro tick".

scanner-darkly:
yeah, agreed it's the most intuitive approach. 

updated the PR so now `M` will allow the current interval to finish before using the new one.

scanner-darkly:
@tehn this is ready for a review when you get a chance!

tehn (approved):
i've been following along, this seems correct to me!

PR #293, opened 2022-10-18
Fix CROW_Q3 operation

beels:
#### What does this PR do?

Corrects a typo that prevents the `CROW.Q3` op from working.

Connect crow to teletype via i2c; set up a crow event handler that responds to `CROW.Q3` and produces visible output in druid; invoke `CROW.Q3` on teletype; observe that the event handler is called with the PR, but not on the unpatched main branch.

#### I have,
* [ ] run tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v4.0.x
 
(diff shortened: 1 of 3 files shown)


scanner-darkly:
looks good! could you add a line about the fix to changelog and whats_new?

scanner-darkly (approved):

PR #306, opened 2023-04-10
Accept all line ending styles for USB load

### LINKED ISSUE #104: scenes with `#` in description don't load properly

scanner-darkly:
#### What does this PR do?

previously, only *nix style line endings (LF) were accepted. i've updated deserialization code to accept DOS (CRLF) and Mac (CR) line ending styles as well. i also used it as an opportunity to make the deserialization code more robust.

as a consequence of this update, now it will also ignore any lines that start with `#` and are not followed by M, I, P, G or 1-8, which means script files can be annotated with comments. it also effectively fixes #104. 

while working on this, i realized that we use magic numbers for script numbers in multiple places, which makes it more difficult to create firmware mods for increased script count, so i've addressed that as well (i'm also considering revisiting the idea of shadow scripts..)

finally, as there was no need for a dedicated enum for script number, i've changed it to uint8_t and this surfaced an issue in our tests where the script number wasn't set properly, and a couple of places in the main code where we weren't doing boundary checks, so i fixed that as well.

save/load scenes to USB using different line endings.

#104 

#### I have,

* [-] updated the documentation
* [-] updated `help_mode.c` (if applicable)
* [x] run tests


tehn (approved):
this is an excellent cleanup, thank you so much for this level of attention. the code feels much better.

PR #310, opened 2023-04-13
improve TR.P accuracy

scanner-darkly:
#### What does this PR do?

previously, TR pulses were processed by `tele_tick` which gets triggered by a 10ms timer, which resulted in the actual pulse length being off by up to 9ms. this PR introduces dedicated timers for each TR output to improve the accuracy. as there are only 4 trigger outputs, i don't think adding 4 timers should make any impact.

additionally, i fixed a bug i discovered where killing delays would not restore TR outputs that had a pulse in progress.

i tested this with a scope and confirmed that the pulse time matches the one set with `TR.TIME`. also confirmed TR pulses work in general - triggering different outputs, changing the pulse time while a pulse is in progress, killing delays.

#### I have,
* [x] run tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@
 - **FIX**: support all line ending types for USB load
 - **FIX**: fix `STATE` not accounting for `DEVICE.FLIP`
 - **FIX**: fix MIDI IN ops channel number being off by 1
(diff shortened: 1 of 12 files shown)


tehn (approved):
awesome!!

PR #313, opened 2023-04-30
new ops to execute script lines and execute scripts as functions

scanner-darkly:
#### What does this PR do?

this PR adds the following:
- the ability to execute a script as a function so that it can be used in an expression
- the ability to execute a specific script line (also can be used as a function)

additional op variations allow passing in 1 or 2 parameters. there are dedicated ops to get the parameters and set the return value. there are also extra ops that execute a line within the same script.

the complete list of added ops:

`$F script` - execute a script as a function
`$F1 script param1` - execute a script with 1 param
`$F2 script param1 param2` - execute a script with 2 params

`$L script line` - execute a script line as a function
`$L1 script line param1` - execute a script line with 1 param
`$L2 script line param2` - execute a script line with 2 params

`$S line` - execute a script line as a function within the same script
`$S1 line param1` - execute a script line within the same script with 1 param
`$S2 line param2` - execute a script line within the same script with 2 params

`I1` - get first parameter
`I2` - get second parameter 
`FR` - get/set the return value

i've also identified and fixed a bug where when a script was executed in some scenarios and the maximum depth of 8 scripts was already reached, we were removing execution state from stack even though it wasn't pushed first, which could result in subtle errors.

finally, i've updated the warning message shown when the flash memory is fresh as several people were confused by it, and increased the time for the message to be displayed to 10 sec.

#### I have,
* [x] run tests


tehn (approved):
fascinated to see where this leads TT scripting!

code looks good, haven't tested on hardware.

PR #321, opened 2023-05-07
Usb minimal menu

Dewb:
#### What does this PR do?

Adds a minimal menu for USB operations. This feature design was suggested by @beels as a minimal useful gesture towards his 
WIP [usb-disk-menu branch](https://github.com/beels/teletype/tree/feature/usb-disk-menu).

Insert a USB stick. You should see a menu with four options. Turn the param knob to highlight an option and press the front button to select it. `EXIT` will return to live mode without doing anything. `WRITE TO USB` will copy scenes from flash to the `tt00s.txt` files. `READ FROM USB` will copy the contents of the `tt00.txt` files to flash memory. `DO BOTH` will perform both a write and then a read, the previous behavior.

From Discord discussion, there are a couple motivations here: 
* providing a failsafe to avoid inadvertent usb operations if an inserted USB drive is physically touched in such a way that it retriggers the MSC connection event.
* making a small gesture in this direction to reduce the activation energy for continuing to enhance the usb menu in future releases.

#### I have,

### DIFF
--- a/module/globals.h
+++ b/module/globals.h
@@ -40,6 +40,8 @@ void set_mode(tele_mode_t mode);
 void set_last_mode(void);
 void clear_delays_and_slews(scene_state_t *ss);
 
+void assign_main_event_handlers(void);
+
 // global copy buffer
 extern char copy_buffer[SCENE_TEXT_LINES][SCENE_TEXT_CHARS];
 extern uint8_t copy_buffer_len;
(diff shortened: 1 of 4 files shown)


Dewb:
recommend viewing the diff in ignore-whitespace mode.

tehn (approved):
thanks! this is a huge UI improvement.

code looks good, did not test on hardware.

scanner-darkly:
looks good!

Issue #145, opened 2018-05-05
CV output calibration


tehn:
in the case of CV channels not matching it'd be helpful to have a command that stores a calibration value to flash (and is automatically used in future bootups)

ie `CV.CAL 1 100` which has the effect of an additional offset.

gssjr:
+1

dndrks:
+1 via support

Dewb:
I have ann [incomplete implementation](https://github.com/dewb/teletype/tree/dewb/calibration) of the calibration op strategy [proposed here](https://llllllll.co/t/teletype-cv-calibration/5654/21); I got stuck on deciding exactly where the calibration data should be stored.

PR #322, opened 2023-05-11
Allow calibration of CV outputs with CV.CAL and CV.CAL.RESET

Dewb:
#### What does this PR do?

Adds a CV.CAL op to create calibration scaling parameters for a CV output and store them in flash memory, and a CV.CAL.RESET op to restore the scaling to the default.

Using the op to calibrate requires a digital multimeter with at least millivolt precision. The procedure is:

1. run `L 1 4: CV I V 1` and record measurements to 3 decimal places with a good digital multimeter.
2. run `L 1 4: CV I V 3` and record measurements to 3 decimal places etc. (second column)
3. run `CV.CAL <output num> <V 1 measurement in millivolts> <V 3 measurement in millivolts>` for each output.

For example, I measure the following on my unit:

So with that “before” data I enter:
```
CV.CAL 1 989 2982
CV.CAL 2 1017 3011
CV.CAL 3 1002 2997
CV.CAL 4 996 2984
```

and afterwards I observe the values reported in the “after” columns above.

https://llllllll.co/t/teletype-cv-calibration/5654

Follow the calibration procedure described above and measure the before and after states. Power off the module and turn it back on and Re measure to confirm that the calibration was saved and is still being applied. 

For a couple years now I have been using a patch to do calibration with floating-point math and hardcoded constants (as described in the lines thread), and I finally got around to rewriting it using fixed point integer math and implementing an op for it so I could more easily use the 5.0 beta branch for an upcoming performance.

#145 

#### I have,


Dewb:
This still needs broader testing (only tried on my unit so far) and some additional tweaks (right now you need to recalibrate if you toggle `DEVICE.FLIP`, that should be fixable.) But I wanted to see if there were any concerns about the overall approach.

tehn:
oop, too fast on the button pushing. didn't mean to un-draft. just wanted to comment and say the code looks good and this is a great feature addition.

scanner-darkly:
the code looks good!

still not convinced recalibration would be required after `DEVICE.FLIP` - unless i'm missing something, you just need to ensure you tie the calibration data based on the actual indexes, so you just need to check for `device_config.flip` when storing it, and then in `cvTimer_callback` you apply based on the actual indexes as well, you don't even need to check for `device_config.flip` there.

curious why the calibration data is stored with a scene - would you ever have a scenario where it would be different between different scenes? also wouldn't it mean that your calibration could be off if you load a scene that contains different calibration values?

Dewb:
Oh yes, orientation-independent calibration should be trivially implementable exactly as you describe. I just haven't had time to do it yet.

> curious why the calibration data is stored with a scene

I had the same question initially when I started looking at how the existing calibration data is stored. It is persisted in its own area of flash memory, the in-memory object is just cached on the scene state (and copied between scenes when you switch, see for example https://github.com/monome/teletype/blob/33f72cbb4bdbf1538209d044b4a454793ed74f18/src/ops/init.c#L61 .)

src/scale.c
```diff
@@ -2,11 +2,28 @@
...
     16383,
-    { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 },
-    { 16383, 16383, 16383, 16383, 16383, 16383, 16383, 16383, 16383, 16383,
-      16383, 16383, 16383, 16383, 16383, 16383 }
+    // f_min
+    {
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    },
```
Dewb, line 16: I believe these default f_min and f_max tables should have been expanded in https://github.com/monome/teletype/pull/288

Dewb: Or maybe at this point it would be less wasteful to initialize new `cal_data_t`s with code, and save all this program memory.

scanner-darkly:
yeah it's possible i'm missing some nuance, but i think using the scene state as a cache for something hardware specific just so that it's available to ops is an antipattern - it's counterintuitive, since it creates an impression that it's stored with the scenes, and main.c should encapsulate all the hardware specific stuff and expose getters/setters for anything that needs to be controlled by ops (which is how device flip is handled).

i don't think it needs to be refactored though, unless you feel like tackling it!

Dewb:
Fixed the `DEVICE.FLIP` issue, calibration now applies correctly regardless of the flip state. I also replaced `blank_cal_data` with a function and documented the new ops. I do think the cal_data_t access should eventually move from scene state to device_config, but I think that's a task for a follow-up PR. This should be ready to go!

src/scale.c
Dewb: (done)

Dewb:
Oops, I spoke too soon; I needed to add `DEVICE.FLIP` to the calibration OP in addition to the CV callback.

In the process, I noticed a couple of other things: one, `CV.GET` was not handling `DEVICE.FLIP` correctly, `aout` is indexed with the software output number. 

The other is maybe a little more controversial: `DEVICE.FLIP` should really be refreshing the CV and TR outputs. I can see a case for not disturbing the current state after a flip until the outputs are touched, but it leaves the module in a weird state. `TR n`, `CV n`, and `CV.GET n` will report values that don't match the hardware, while TR pulses in progress will be applied to their new locations, leaving the original location unchanged. Things will be in an inconsistent state until all TR and CV outputs are updated in the new ordering, so we might as well do that immediately when the flip state is changed.

module/main.c
```diff
@@ -1101,7 +1111,18 @@ void tele_cv_off(uint8_t i, int16_t v) {
 }
 
 uint16_t tele_get_cv(uint8_t i) {
-    return aout[(device_config.flip ? 3 - i : i)].now;
```
Dewb, line 1104: Here's the CV.GET fix.

module/main.c
```diff
@@ -1101,7 +1111,18 @@ void tele_cv_off(uint8_t i, int16_t v) {
...
+    return aout[i].now;
+}
```
Dewb, line 1115: Here's the CV.GET fix, unrelated to calibration.

scanner-darkly (approved):

scanner-darkly:
thank you for fixing `CV.GET`! and good catch on `DEVICE.FLIP` not updating the outputs - although i'm not convinced about its utility as a musical gesture :) a more useful variation could be being able to freely assign physical outputs to indices..

PR #323, opened 2023-07-15
Correctly document N.CS scales 7 & 8

Dewb:
#### What does this PR do?

Adjusts the documentation for the N.CS scales to match the implementation in [table.c](https://github.com/monome/teletype/blob/7041e67742a84e93c4f0d2797348e00d5a76dd99/src/table.c#L135)

N.CS introduced here: https://llllllll.co/t/teletype-firmware-discussion/13961/49

Compare the documentation to the results of the N.CS op on scales 7 and 8.

I noticed the issue when I was comparing the N.CS implementation to the similar norns functionality in https://github.com/monome/norns/pull/1688.

(n/a)

#### I have,

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - **NEW**: new op: `CV.GET`
 - **NEW**: basic menu for reading/writing scenes when a USB stick is inserted
 - **NEW**: new ops: `CV.CAL` and `CV.CAL.RESET` to calibrate CV outputs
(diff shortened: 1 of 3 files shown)


Dewb:
help_mode.c was already correct. With these changes, the docs should reflect the actual behavior.

Unfortunately there's also an implementation issue: it looks like the order of scales 7 and 8 is swapped between N.S and N.CS. Not sure if it's worth breaking the current behavior just to make it more consistent.

PR #336, opened 2024-11-01
Fix potential crash in help mode reverse search

Dewb:
#### What does this PR do?

It's possible for help mode reverse search to index beyond the bounds of a page. On hardware, this is likely not a hard crash, but it could lead to unexpected behavior. On VCV Rack hitting this results in a crash (see https://github.com/Dewb/monome-rack/issues/206).

See steps to reproduce in https://github.com/Dewb/monome-rack/issues/206

#### I have,
* [X] updated `help_mode.c` (if applicable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v5.0.x:
+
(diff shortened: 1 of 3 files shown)


tehn (approved):
thanks!

